PR #22, opened 2015-02-03
ISPN-5200 Enhance performance of annotation driven marshallers by cachin...

anistor:
...g nested entity marshallers instead of always looking them up

Jira: https://issues.jboss.org/browse/ISPN-5200


anistor:
Integrated in master.

PR #23, opened 2015-02-04
ISPN-5203 Protostream annotation attributes that do not have a default value should be required

anistor:
Jira: https://issues.jboss.org/browse/ISPN-5203


anistor:
Integrated in master.

PR #29, opened 2017-07-20
IPROTO-32 Implement conversion from JSON to protobuf

gustavonalle:
https://issues.jboss.org/browse/IPROTO-32

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
+++ b/core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
@@ -1,30 +1,60 @@
 package org.infinispan.protostream;
 
+import static com.google.gson.stream.JsonToken.END_DOCUMENT;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_BOOL;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_BYTES;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_DESCRIPTOR_FULL_NAME;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_DESCRIPTOR_ID;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_DOUBLE;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_ENUM;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_FIXED32;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_FIXED64;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_FLOAT;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_INT32;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_INT64;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_MESSAGE;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_SFIXED32;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_SFIXED64;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_SINT32;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_SINT64;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_STRING;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_UINT32;
+import static org.infinispan.protostream.WrappedMessage.WRAPPED_UINT64;
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.Reader;
 import java.nio.ByteBuffer;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Base64;
 import java.util.Date;
 import java.util.GregorianCalendar;
+import java.util.Set;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
```diff
@@ -1,31 +1,62 @@
...
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.Reader;
 import java.nio.ByteBuffer;
 import java.text.DateFormat;
+import java.text.ParseException;
```
anistor: Not used.

core/src/test/java/org/infinispan/protostream/domain/Account.java
```diff
@@ -102,15 +126,33 @@ public void setCurrencies(Currency[] currencies) {
...
    public String toString() {
       return "Account{" +
             "id=" + id +
             ", description='" + description + '\'' +
             ", creationDate='" + creationDate + '\'' +
             ", limits=" + limits +
-            ", blurb=" + blurb +
-            ", currencies=" + Arrays.toString(currencies) +
+            ", blurb=" + blurb.stream().map(Arrays::toString).collect(Collectors.toList()) +
             ", unknownFieldSet='" + unknownFieldSet + '\'' +
```
anistor: I'd prefer to have the unknownFieldSet as the last printed stuff.

core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
```diff
@@ -166,6 +202,388 @@ public static String toCanonicalJSON(ImmutableSerializationContext ctx, byte[] b
...
+               case END_DOCUMENT:
+                  break;
+               default:
+                  throw new IllegalStateException("Invalid top level object! Found token: " + token);
+            }
+         }
+         writer.flush();
+         return baos.toByteArray();
+      } catch (MalformedJsonException e) {
+         throw new IllegalStateException("Invalid Json");
```
anistor: Let's propagate the cause here. Also, I believe ```JSON``` is the correct term, not Json.

gustavonalle:
Updated

core/src/main/java/org/infinispan/protostream/TagHandler.java
```diff
@@ -13,8 +15,10 @@
 
    /**
     * Start of top level message. Do whatever required init here.
+    *
+    * @param descriptor The {@link GenericDescriptor} for the top level Enum or Message. For primitives it is null.
     */
-   default void onStart() {
+   default void onStart(AnnotatedDescriptor descriptor) {
```
gustavonalle: I've added this in order to have access to the full type name in the beginning of the stream in order to render the ```_type``` field in the JSON message. Do you see any issue with backwards compat? Protostream is not usually a user face library, and to migrate ISPN it's necessary to replace ```onStart()``` to ```onStart(null)``` in 2 or 3 places

anistor: This will not be a compat issue. This change will go in protostream 4.2 which will only be used by infinispan 9.2+

anistor: I'd prefer the ```descriptor``` arg to be a ```GenericDescriptor```. There's no need to for this one to be a field descriptor too. I know there is one occurrence where this seems to be needed but that is a mistake. I'll comment on it.

gustavonalle: Regarding compat, I was thinking about backporting to JDG

anistor: Same for jdg. All jdg versions previous to the upcoming 7.2 will be stuck with a pre 4.2 version of protostream. Only jdg 7.2 will start using 4.2 and will need to adapt to this change.

gustavonalle: My concern is whether source compatibility would be maintained when migrating from 7.1 to 7.2, but since this API is internal, users should not be affected, unless they are using protostream to other ends

gustavonalle: > I'd prefer the descriptor arg to be a GenericDescriptor

np, I need an broader descriptor for the cases where there is no message being rendered

core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
```diff
@@ -142,6 +175,8 @@ public static ByteBuffer toWrappedByteBuffer(ImmutableSerializationContext ctx,
 
       boolean isFirstField = true;
 
+      boolean isTypeMetaField = false;
```
anistor: I don't think we need ```isTypeMetaField```. The logic for writing the _type field only once and adding a proper coma after it can be also done using ```isFirstField```

core/src/main/java/org/infinispan/protostream/TagHandler.java
anistor: I gave some more thought about AnnotationDescriptor vs GenericDescriptor in onStart method. Going with AnnotationDescriptor seems a bit easier but is not very orthogonal and also does not adhere to the initial javadoc of TagHandler.onStart  :)  So let's go with GenericDescriptor :)

I've done this change and pushed a branch here https://github.com/anistor/protostream/commits/t_gustavonalle_json   (see last commit).

gustavonalle:
@anistor Updated with your commit

anistor:
thanks! was hoping you would squash them.

gustavonalle:
I can squash, one sec

anistor:
Integrated in master. Thanks @gustavonalle !

PR #37, opened 2018-06-18
Support for Dynamic Entities

gustavonalle:


gustavonalle:
This proposal has a limitation: it will work as long as a single marshaller provider is used.
Proper support for dynamic entities will come in the next version, so closing this...

PR #40, opened 2018-07-05
IPROTO-64 Annotation based marshallers should unmarshall empty repeat…

anistor:
…able fields as empty collections instead of null

https://issues.jboss.org/browse/IPROTO-64


anistor:
Integrated in master.

PR #44, opened 2018-08-09
IPROTO-67 Remove the class parameter from ProtoStreamWriter.writeEnum

anistor:
* a deprecated method having a class parameter is still kept for backward compat

https://issues.jboss.org/browse/IPROTO-67


anistor:
Integrated in 4.2.x and master.

PR #47, opened 2018-08-22
IPROTO-71 Fix compat issue with JDK 9,10,11 and add CI testing for them

anistor:
* Build with JDK 8,9,10,11; both Oracle and OpenJDK are used

https://issues.jboss.org/browse/IPROTO-71


anistor:
Integrated in master.

PR #63, opened 2019-03-18
[4.2.x] IPROTO-90 NPE when writing message having a null field of a boxed pri…

anistor:
…mitive type that is also marked required

* this happens with annotations

https://issues.jboss.org/browse/IPROTO-90


anistor:
Integrated in 4.2.x.

PR #74, opened 2019-07-29
IPROTO-20 Should not be able to use the numbers 19000 though 19999 fo…

anistor:
…r fields

https://issues.jboss.org/browse/IPROTO-20


anistor:
Merged in master (4.3.x).

PR #84, opened 2019-11-22
IPROTO-122 JDK version check in the annotation processor fails on Ope…

anistor:
…njdk 13

https://issues.jboss.org/browse/IPROTO-122


anistor:
Merged in master (to be released with 4.3.1.Final)

PR #88, opened 2020-01-23
IPROTO-126 JDK version check in the annotation processor fails on JDK…

anistor:
… 14-ea / 15-ea

https://issues.redhat.com/browse/IPROTO-126


anistor:
merged in master

PR #91, opened 2020-03-05
IPROTO-132 Enum classes with abstract methods don't compile

ryanemerson:
https://issues.redhat.com/browse/IPROTO-132

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/annotations/impl/ProtoMessageTypeMetadata.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/impl/ProtoMessageTypeMetadata.java
@@ -339,7 +339,7 @@ public final class ProtoMessageTypeMetadata extends ProtoTypeMetadata {
                   // MESSAGE is the default and stands for 'undefined', we can override it with a better default
                   protobufType = Type.BYTES;
                }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/protostream/annotations/impl/ProtoMessageTypeMetadata.java
```diff
@@ -339,7 +339,7 @@ private void discoverFields(XClass clazz, Set<XClass> examinedClasses) {
                   // MESSAGE is the default and stands for 'undefined', we can override it with a better default
                   protobufType = Type.BYTES;
                }
-               if (!javaType.isArray() && !javaType.isPrimitive() && javaType.isAbstract()) {
+               if (!javaType.isArray() && !javaType.isPrimitive() && (javaType.isAbstract() && !javaType.isEnum())) {
```
anistor: OMG, unneeded grouping with ( ) !!

I think this fix is needed on line 493 also, but no existing test highlights that case.

ryanemerson: I've updated `Simple.java` so that the method variation is also covered by the testsuite.

anistor:
Merged. Thanks @ryanemerson !

PR #92, opened 2020-03-17
Integrate forbiddenapis + fix various forbiddenapi usage.

johnou:
Motivation:

Checks Java byte code against a list of "forbidden" API signatures.

Modification:

Integrate forbiddenapis plugin and fix usages.

Result:

Future proof code.

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/annotations/impl/types/ReflectionClassFactory.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/impl/types/ReflectionClassFactory.java
@@ -19,6 +19,7 @@ import java.util.Optional;
 import javax.lang.model.type.TypeMirror;
 
 import org.infinispan.protostream.annotations.ProtoDoc;
+import org.infinispan.protostream.util.internal.SuppressForbidden;
 
 /**
  * Implementation relying on reflection.
@@ -86,6 +87,7 @@ public final class ReflectionClassFactory implements XTypeFactory {
       return null;
    }
 
+   @SuppressForbidden(reason = "java.lang.reflect.AccessibleObject#setAccessible(boolean)")
    private final class ReflectionClass implements XClass {
 
       private final Class<?> clazz;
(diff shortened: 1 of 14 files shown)


johnou:
@anistor let me know if you would prefer that I split this into multiple pull requests.

johnou:
@anistor this one is ready to go too, if I work out an elegant solution to IPROTO-88 I will uncomment the custom forbidden API..

core/src/main/java/org/infinispan/protostream/annotations/impl/types/ReflectionClassFactory.java
```diff
@@ -109,9 +111,9 @@ public XClass fromTypeMirror(TypeMirror typeMirror) {
                   try {
                      f.setAccessible(true);
                      e = (Enum<?>) f.get(clazz);
-                  } catch (IllegalAccessException iae) {
+                  } catch (IllegalArgumentException | IllegalAccessException ex) {
```
anistor: I'd rather not handle IllegalArgumentException or any other exceptions here, just let it explode. But IllegalAccessException is not a runtime exception.

anistor:
What is the purpose of the locale related fixes @johnou ? Do we have any failures?

core/src/main/java/org/infinispan/protostream/annotations/impl/types/ReflectionClassFactory.java
johnou: Got it, I'll roll it back. Point was to throw an exception with useful information.

johnou:
@anistor no failures, could technically suppress the check for those if you prefer since there should be no problems with english only messages - let me know what you prefer.

core/src/main/java/org/infinispan/protostream/annotations/impl/types/ReflectionClassFactory.java
johnou: reverted.

akkarapon26 (approved):

johnou:
let's close this.

PR #105, opened 2020-09-21
IPROTO-148 Add ProtoField.value() and ProtoEnumValue.value() as a con…

anistor:
…venient alias for number()

https://issues.redhat.com/browse/IPROTO-148


anistor:
merged in master

PR #111, opened 2020-10-13
PROTO-167 Allow setter methods to be annotated with @ProtoField even …

anistor:
…when @ProtoFactory is used

* this was previously not allowed because we tried to not allow existence of setters at all
   and make all @ProtoFactory created objects immutable

https://issues.redhat.com/browse/IPROTO-167


anistor:
merged in master and 4.3.x

PR #114, opened 2020-10-19
IPROTO-137 Support Schema and Marshaller generation for external classes

anistor:
* introduce ProtoBridgeFor annotation to create annotation based
  marshallers for peer classes that cannot be annotated

https://issues.redhat.com/browse/IPROTO-137

NOTE: ```ProtoBridgeFor```used to be named ```ProtoMarshallerFor``` until just now. I'm open for other name suggestions. Since the next release is an Alpha we can have it as is and rename it sometime before final.


anistor:
Enums are not handled here. See https://issues.redhat.com/browse/IPROTO-171

anistor:
merged in master

PR #120, opened 2020-12-18
IPROTO-178 Canonical json to protobuf conversion writes TypeId as int…

anistor:
…32 instead of uint32

* IPROTO-116 changed type id from int32 to uint32 but the change
 was not correctly propagated in all places

https://issues.redhat.com/browse/IPROTO-178


anistor:
merged in master and 4.3.x

PR #124, opened 2021-03-02
IPROTO-56 Support for dynamic models

anistor:
https://issues.redhat.com/browse/IPROTO-56

The approach is very similar to Gustavo's initial proposal but without the limitations ( https://github.com/infinispan/protostream/pull/37 )


anistor:
Merged in master for now because today we'll have a quick release. Eventual changes based on feedback will be done later.

PR #134, opened 2021-04-09
IPROTO-208 Primitive types in proto files should not be case insensitive

anistor:
* Type.valueOf(typeName.toUpperCase()) was a huge mistake

https://issues.redhat.com/browse/IPROTO-208


anistor:
merged in master and 4.3.x

PR #135, opened 2021-04-13
Added buffer access to TagReader

shuawest:
Exposing the bufferArray/InputStream on the TagReader interface will open the ability to use alternative frameworks for reading from infinispan, such as protobuf-java. Writing is already possible. For applications with many complex models, this will prevent the need to rebind to the protostream specific classes, and push/pull protobuf straight from other sources into the cache/grid. For large objects, the InputStream was also exposed.

The limitation is that proto messages in a single key:value pair can't mix accessing the full buffer array, with a protostream customer marshaller, nor protostream annotated auto objects. The entire payload of the WrappedMessage is read. Checks were added to the enhancement to prevent mixing, and make it clear to the user through an IllegalStateException, and comments on the TagReader interface. 

Tests were added for the happy path with the byte[] and InputStream, and a failure test for the mixing of marshalling modes. 

With this addition, a client can create a generic protobuf marshaller for existing protobuf-java classes using [1] and a context initializer like [2]. 

[1] 
```
public class GenericProtobufMarshaller<T extends AbstractMessage> implements ProtobufTagMarshaller<T> {
    private static final transient Logger log = LoggerFactory.getLogger(GenericProtobufMarshaller.class);

    private final Class<? extends T> javaClass;
    private final String typeName;
    private final ByteArrayParseFunction<byte[], T> parseFromMethod;

    public GenericProtobufMarshaller(
        final Class<? extends T> javaClass, 
        final String typeName, 
        ByteArrayParseFunction<byte[], T> parseFromMethod) 
    {
        this.javaClass = javaClass;
        this.typeName = typeName;
        this.parseFromMethod = parseFromMethod;
    }

    @Override
    public Class<? extends T> getJavaClass() {
        return javaClass;
    }

    @Override
    public String getTypeName() {
        return typeName;
    }

    @Override
    public T read(ReadContext ctx) throws IOException {
        TagReader reader = ctx.getReader();
                       
        byte[] rawBytes = reader.fullBufferArray();
        T t = parseFromMethod.apply(rawBytes);
        
        return t;
    }

    @Override
    public void write(WriteContext ctx, T t) throws IOException {
        byte[] tBytes = t.toByteArray();
        log.info("writing bytes: {}", tBytes);

        TagWriter out = ctx.getWriter();
        out.writeRawBytes(tBytes, 0, tBytes.length);
    }

    @FunctionalInterface
    public interface ByteArrayParseFunction<T, R> {
        R apply(T rawBytes) throws InvalidProtocolBufferException;
    }

}
```

[2]
```
public class MySerializationContextInitializer implements SerializationContextInitializer {
...
   @Override
    public void registerMarshallers(SerializationContext ctx) {
        ctx.registerMarshaller(new GenericProtobufMarshaller<>(Payment.class, "realtimepayments.Payment", Payment::parseFrom));  
        ctx.registerMarshaller(new GenericProtobufMarshaller<>(Payer.class, "realtimepayments.Payer", Payer::parseFrom));          
        ctx.registerMarshaller(new GenericProtobufMarshaller<>(Payee.class, "realtimepayments.Payee", Payee::parseFrom));       
        ctx.registerMarshaller(new GenericProtobufMarshaller<>(Transaction.class, "realtimepayments.Transaction", Transaction::parseFrom));
    }
...
}
```

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/TagReader.java
+++ b/core/src/main/java/org/infinispan/protostream/TagReader.java
@@ -1,6 +1,7 @@
 package org.infinispan.protostream;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.ByteBuffer;
 
 /**
@@ -84,4 +85,25 @@ public interface TagReader extends RawProtoStreamReader {
     * Returns back to a previous limit returned by pushLimit.
     */
    void popLimit(int oldLimit);
+
+   /** 
+    * Returns the full buffer array allowing an alternative protobuf parser to be used in marshallers,
+    * and at the root message level. This should not be mixed with the other tag reader read***() 
+    * methods, or else an IllegalStateException will be thrown. Therefore you cannot mix protostream
+    * annotated models with other parsers reading the raw payload array. 
+    */
+   byte[] fullBufferArray() throws IOException;
+
+   /** 
+    * Returns the input stream allowing an alternative protobuf parser to be used in marshallers,
+    * and at the root message level. This should not be mixed with the other tag reader read***() 
+    * methods, or else an IllegalStateException will be thrown. Therefore you cannot mix protostream
+    * annotated models with other parsers reading the raw payload input stream.  
+    */
+   InputStream fullBufferInputStream() throws IOException;
+
+   /**
+    * @return Returns true if the original payload is InputStream based.
+    */
+   boolean isInputStream(); 
 }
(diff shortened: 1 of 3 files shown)


tristantarrant:
Thanks @shuawest , I've pushed it as https://github.com/infinispan/protostream/commit/0c28165273abe0d51e3d888fad8ab61c283634c2 fixing the conflict and adding a Jira

PR #151, opened 2022-04-07
IPROTO-229 Use String.getBytes() to get an UTF-8 encoded byte array

pruivo:
https://issues.redhat.com/browse/IPROTO-229


tristantarrant:
Merged, thanks

PR #157, opened 2022-04-25
IPROTO-223 MarshallerByteCodeGenerator - NoSuchMethodError Class.getM…

wburns:
…odule when running tests with JDK8

https://issues.redhat.com/browse/IPROTO-233


tristantarrant:
While this fixes JDK 8, it breaks JDK 11 and above

tristantarrant:
The fix should check for the JVM version, not the version of the class file. Fixed by https://github.com/infinispan/protostream/pull/158

PR #168, opened 2022-07-29
IPROTO-239 Use correct generated class

fax4ever:
https://issues.redhat.com/browse/IPROTO-239

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -104,12 +104,6 @@
         </dependency>
 
         <dependency>
-            <groupId>javax.annotation</groupId>
-            <artifactId>javax.annotation-api</artifactId>
-            <optional>true</optional>
-        </dependency>
-
-        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>sample-domain-definition</artifactId>
             <scope>test</scope>
--- a/integrationtests/pom.xml
+++ b/integrationtests/pom.xml
@@ -38,11 +38,6 @@
         </dependency>
 
         <dependency>
-            <groupId>javax.annotation</groupId>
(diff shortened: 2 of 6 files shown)


pruivo:
@fax4ever the minimum JDK version is 11. Can't we just use the JDK9+ annotation all the time (and probably update the annotation-api dependency)?

gsmet:
@pruivo we talked about it with Katia but the README seems to indicate you still support a JDK 8 runtime for protostream: `Target runtime platform is Java 8.`. But maybe it hasn't been updated.

gsmet (approved):
Should fix the issue we have with Quarkus, thanks!

pruivo:
The README isn't updated yet (we forgot). 
We changed the baseline to JDK 11 a while back for Protostream 4.5+
https://github.com/infinispan/protostream/blob/93799230b972131294796d41b5c2d33a79aa0da1/parent/pom.xml#L103-L104

gsmet:
OK, well, in this case, we can just switch to the new annotation. That will simplify things for sure :).

And maybe update the README too.

fax4ever:
Thanks @pruivo and @gsmet for the review.
I pushed the change to remove the annotation-api dependency, we don't need it anymore with JDK 11+.
I put aside the patch to support both JDK 8- and 11+, maybe we can use it if we need to backport the fix to some old ProtoStream version.

tristantarrant:
Merged, thanks

fax4ever:
Thank you!

gsmet:
Thanks!

gsmet:
@fax4ever @tristantarrant @pruivo it would be awesome if you could include this in the next Infinispan 14 dev release!

PR #39, opened 2018-06-28
IPROTO-62 Implement annotation based schema generation support for 'oneof'

anistor:
https://issues.jboss.org/browse/IPROTO-62

PR #169, opened 2022-08-02
IPROTO-62 Implement annotation based schema generation support for 'oneof'

fax4ever:
https://issues.redhat.com/browse/IPROTO-62

I think that we can evaluate the idea of merge this Adrian's contribution to support the generation of `oneof` clause from `@ProtoField` annotations


fax4ever:
I opened this PR, but probably it is better to merge the one Adrian created to solve the same issue, that has the exact same code of this one: https://github.com/infinispan/protostream/pull/39

PR #191, opened 2023-05-05
IPROTO-264 Remove stream buffer when unmarshalling objects

wburns:
https://issues.redhat.com/browse/IPROTO-264

Reduces each invocation by 100-200 ns for relatively small objects

Ran from https://github.com/infinispan/infinispan-benchmarks/blob/main/protostream/src/main/java/org/infinispan/protostream/ProtostreamBenchmark.java

4.6.3-SNAPSHOT
```
Benchmark                                    (byteArrayOrStream)  Mode  Cnt     Score   Error  Units
ProtostreamBenchmark.testUnmarshallAddress                 false  avgt    6   340.235 ± 1.417  ns/op
ProtostreamBenchmark.testUnmarshallMetadata                false  avgt    6   537.461 ± 2.175  ns/op
ProtostreamBenchmark.testUnmarshallUser                    false  avgt    6  1057.641 ± 6.004  ns/op
```
4.6.2.Final
```
Benchmark                                    (byteArrayOrStream)  Mode  Cnt     Score   Error  Units
ProtostreamBenchmark.testUnmarshallAddress                 false  avgt    6   528.840 ± 3.347  ns/op
ProtostreamBenchmark.testUnmarshallMetadata                false  avgt    6   694.986 ± 2.925  ns/op
ProtostreamBenchmark.testUnmarshallUser                    false  avgt    6  1159.652 ± 4.904  ns/op
```

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
@@ -7,20 +7,18 @@ import static org.infinispan.protostream.descriptors.WireType.MAX_VARINT_SIZE;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PushbackInputStream;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 import org.infinispan.protostream.ImmutableSerializationContext;
 import org.infinispan.protostream.MalformedProtobufException;
 import org.infinispan.protostream.ProtobufTagMarshaller;
-import org.infinispan.protostream.ProtobufUtil;
 import org.infinispan.protostream.TagReader;
 import org.infinispan.protostream.descriptors.WireType;
 
@@ -63,7 +61,7 @@ public final class TagReaderImpl implements TagReader, ProtobufTagMarshaller.Rea
    }
 
    public static TagReaderImpl newNestedInstance(ProtobufTagMarshaller.ReadContext parent, InputStream input) {
-      return new TagReaderImpl((TagReaderImpl) parent, new InputStreamDecoder(input, ProtobufUtil.DEFAULT_STREAM_BUFFER_SIZE));
+      return new TagReaderImpl((TagReaderImpl) parent, new InputStreamDecoder(input));
    }
 
    public static TagReaderImpl newNestedInstance(ProtobufTagMarshaller.ReadContext parent, byte[] buf) {
@@ -71,7 +69,7 @@ public final class TagReaderImpl implements TagReader, ProtobufTagMarshaller.Rea
    }
 
    public static TagReaderImpl newInstance(ImmutableSerializationContext serCtx, InputStream input) {
-      return new TagReaderImpl((SerializationContextImpl) serCtx, new InputStreamDecoder(input, ProtobufUtil.DEFAULT_STREAM_BUFFER_SIZE));
+      return new TagReaderImpl((SerializationContextImpl) serCtx, new InputStreamDecoder(input));
    }
 
    public static TagReaderImpl newInstance(ImmutableSerializationContext serCtx, ByteBuffer buf) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
```diff
@@ -846,49 +844,33 @@ private static final class InputStreamDecoder extends Decoder {
...
 
       @Override
       String readString() throws IOException {
          int length = readVarint32();
-         if (length > 0 && length <= end - pos) {
-            String value = new String(buf, pos, length, UTF8);
+         if (length > 0 && length <= limit - pos) {
+            byte[] bytes = readRawByteArray(length);
+            String value = new String(bytes, 0, length, UTF8);
             pos += length;
```
pruivo, line 874: `readRawByteArray` already updated `pos`

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
```diff
@@ -898,51 +880,20 @@ String readString() throws IOException {
...
-         }
-         if (length <= buf.length) {
-            fillBuffer(length);
-            int from = pos;
-            pos += length;
-            return ByteBuffer.wrap(Arrays.copyOfRange(buf, from, pos));
-         }
-         // TODO [anistor] implement a readRawByteBufferLarge, using off-heap allocation
-         return ByteBuffer.wrap(readRawBytesLarge(length));
+         byte[] bytes = readRawByteArray(length);
```
pruivo, line 888: worth checking `length == 0` to avoid creating an empty `ByteBuffer`? 
```java
         if (length == 0) {
            return EMPTY_BUFFER;
         }
```

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
```diff
@@ -1066,186 +995,75 @@ InputStream getInputStream() {
...
-         if (length > 0 && length <= end - pos) {
-            int from = pos;
+         if (length > 0 && length <= limit - pos) {
             pos += length;
-            return Arrays.copyOfRange(buf, from, pos);
+            if (pos > globalLimit) {
+               throw log.globalLimitExceeded();
+            }
+            byte[] array = new byte[length];
+            int readAmount = in.read(array);
```
pruivo, line 1039: shouldn't be `readNBytes()`?
```suggestion
            int readAmount = in.readNBytes(array, 0, length);
```

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
```diff
@@ -1066,186 +995,75 @@ InputStream getInputStream() {
...
-         return bytes;
+         throw log.messageTruncated();
       }
 
       @Override
       protected void skipRawBytes(int length) throws IOException {
-         if (length <= end - pos && length >= 0) {
+         if (length <= limit - pos && length >= 0) {
             pos += length;
+            long skipAmount = in.skip(length);
```
pruivo, line 1058: same question here
```suggestion
            long skipAmount = in.skipNBytes(length);
```

pruivo (changes requested):

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
wburns: No, this way is fine. The length is implied with the provided byte[]. Unless you are thinking something else is wrong?

wburns: Are you thinking because it may return less bytes than it should in a blocking operation?

wburns: I can update both to use that instead to be sure.

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
wburns: Oh the reason I couldn't use this is because it requires Java 12 or newer

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
pruivo: In Infinispan we have all the bytes available, so it shouldn't matter. But using the `readNBytes` seems more correct.

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
pruivo: ah ah, I checked the JDK version of `readNBytes` and I assumed this one was the same :rofl:

core/src/main/java/org/infinispan/protostream/impl/TagReaderImpl.java
```diff
@@ -1066,186 +995,75 @@ InputStream getInputStream() {
...
-         if (requestedBytes + bytesBeforeStart + pos > limit) {
-            // oops, should not exceed the limit
-            return false;
-         }
-
-         // slide existing data if some bytes are already consumed
-         if (pos > 0) {
-            if (end > pos) {
-               System.arraycopy(buf, pos, buf, 0, end - pos);
+         if (in instanceof PushbackInputStream) {
```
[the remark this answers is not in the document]
wburns: Oh weird, I must not have pushed that change. I already have in my working branch.
```java
         if (in.available() > 0) {
            return false;
         }
```

wburns:
I will update the benchmark and rerun the perf tests.

wburns:
Converting to use a repeatable in.read(byte[], int, int) produces way better results.

```java
            while ((readAmount = in.read(array, readTotal, length - readTotal)) != -1) {
               readTotal += readAmount;
               if (readTotal == length) {
                  break;
               }
            }
```

```
Benchmark                                (byteArrayOrStream)  (userByteArraySize)  Mode  Cnt     Score     Error  Units
ProtostreamBenchmark.testUnmarshallUser                false                 8096  avgt    6  2754.104 ± 223.502  ns/op
```

pruivo (approved):

pruivo:
merged! thanks @wburns !

PR #207, opened 2023-10-31
IPROTO-280 Byte and short not parsed into JSON

jabolina:
https://issues.redhat.com/browse/IPROTO-280

The byte/short has a representation in protostream that knows the underlying type. In JSON, it becomes `{"_type":"int32","_value":1}`, but if we parse the JSON back to the protostream representation, the byte array is not the same, as it transforms the value into an int. TBH, I am unsure we can make the byte array the same without breaking compatibility. The problem is that the JSON representation loses the notion of the typing.

### DIFF
--- a/core/src/test/resources/sample_bank_account/bank.proto
+++ b/core/src/test/resources/sample_bank_account/bank.proto
@@ -220,6 +220,39 @@ message Transaction {
    required bool isValid = 9;
 }
 
+message Numerics {
+
+   /**
+    * @Field(store = Store.YES)
+    */
+   optional int32 simpleByte = 1;
+
+   /**
+    * @Field(store = Store.YES)
+    */
+   optional int32 simpleShort = 2;
+
+   /**
+    * @Field(store = Store.YES)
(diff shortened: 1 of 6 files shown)


core/src/test/resources/sample_bank_account/bank.proto
```diff
@@ -220,6 +220,39 @@ message Transaction {
    required bool isValid = 9;
 }
 
+message Numerics {
+
+   /**
+    * @Field(store = Store.YES)
+    */
+   required int32 simpleByte = 1;
```
tristantarrant, line 228: I know this is just a test, but I would prefer if we used `optional` instead of `required` as recommended by the ProtoBuf 3 spec

jabolina: Updated.

tristantarrant:
Merged, thanks. Should we backport to 4.4 ?

jabolina:
I think a backport should be safe. The byte representation does not change.

PR #209, opened 2023-12-21
IPROTO-273 Elements in collections are not properly limited

pruivo:
https://issues.redhat.com/browse/IPROTO-273

Trying this for the second time. There is an option to enable the fix since it changes the binary format. It is disabled by default to keep backward compatibility.


pruivo:
replaced by https://github.com/infinispan/protostream/pull/265

PR #226, opened 2024-01-24
IPROTO-155 Reduce FileDescriptor memory usage

tristantarrant:
https://issues.redhat.com/browse/IPROTO-155

Parse annotations eagerly to avoid keeping them around.

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/descriptors/AnnotatedDescriptor.java
+++ b/core/src/main/java/org/infinispan/protostream/descriptors/AnnotatedDescriptor.java
@@ -1,7 +1,5 @@
 package org.infinispan.protostream.descriptors;
 
-import java.util.Map;
-
 import org.infinispan.protostream.AnnotationParserException;
 
 /**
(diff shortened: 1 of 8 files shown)


core/src/test/java/org/infinispan/protostream/impl/parser/DescriptorsTest.java
```diff
@@ -1154,4 +1153,10 @@ private File asFile(String resourcePath) {
       }
       return null;
    }
+
+   private void calculateSize(Map<String, FileDescriptor> descriptors) {
+      for (Map.Entry<String, FileDescriptor> descriptor : descriptors.entrySet()) {
+         System.out.println("Total Size = " + GraphLayout.parseInstance(descriptor.getValue()).totalSize());
```
pruivo, line 1159: always log this as debug?

core/src/main/java/org/infinispan/protostream/descriptors/AnnotatedDescriptor.java
```diff
@@ -41,17 +39,6 @@ public interface AnnotatedDescriptor {
...
-   /**
-    * Get the documentation annotations in the form of a map of abstract syntax trees of {@link AnnotationElement} nodes.
-    * Only the configured annotations are returned. The unconfigured ones are discarded immediately after parsing.
-    * <p>
-    * The parsing of annotations is performed lazily on first invocation.
-    *
-    * @return the map of annotations
-    * @throws AnnotationParserException if parsing of annotations fails
-    */
-   Map<String, AnnotationElement.Annotation> getAnnotations() throws AnnotationParserException;
```
fax4ever, line 53: Maybe we can keep the method implemented as as default simpy delegating to `getProcessedAnnotation("")`

fax4ever (approved):
Just a suggestion here.
I'm also testing the changes on current Infinispan main.

fax4ever:
Sorry @tristantarrant I only noticed running this patch on current Infinispan main, that there are some cases where the annotations field is filled and `processedAnnotations` not. This probably could be an issue for the remote query code. Full story here => https://infinispan.zulipchat.com/#narrow/stream/118645-infinispan/topic/pr.20queue/near/420641654.

An idea to keep the optimization, is to fill the annotations list only if:

> annotations ⊈ processedAnnotations

at the end of processAnnotations methods. WTDY?

PR #265, opened 2024-04-15
IPROTO-273 Elements in collections are not properly limited

pruivo:
https://issues.redhat.com/browse/IPROTO-273

The third time is a charm... Change the code so the primitive collections/arrays aren't changed.

### DIFF
--- a/types/src/test/java/org/infinispan/protostream/types/java/TypesMarshallingTest.java
+++ b/types/src/test/java/org/infinispan/protostream/types/java/TypesMarshallingTest.java
@@ -4,8 +4,8 @@ import org.infinispan.protostream.GeneratedSchema;
 import org.infinispan.protostream.ImmutableSerializationContext;
 import org.infinispan.protostream.ProtobufUtil;
 import org.infinispan.protostream.SerializationContext;
+import org.infinispan.protostream.config.Configuration;
 import org.infinispan.protostream.impl.Log;
-import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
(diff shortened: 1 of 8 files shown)


pruivo:
@ryanemerson If I recall correctly, you need this to simplify the commands conversation to protostream, right?

types/src/test/java/org/infinispan/protostream/types/java/TypesMarshallingTest.java
```diff
@@ -17,62 +17,127 @@
...
+        assumeTrue(testConfiguration.runTest);
+        if (testConfiguration.isArray) {
+            testConfiguration.method.marshallAndUnmarshallTest(bookArray(), context, true);
+        } else {
+            testConfiguration.method.marshallAndUnmarshallTest(bookCollection(testConfiguration.collectionBuilder), context, false);
+        }
+    }
+
+    @Test
+    public void testPrimitiveCollectionCompatibility() throws IOException {
```
pruivo, line 102: @ryanemerson I added a test here, so I'm confident it is backward compatible.

ryanemerson: Great work! So I guess the idea would be in Infinispan that we will need to read old store entries with `wrapCollectionElements(false)` and migrate the data to use `wrapCollectionElements(true)` on initial server startup?

pruivo: You can read the old store entries with `wrapCollectionElements(true)`. Collections in object fields are not affected by this bug. 
The only issue is if the user has a `RemoteCache<String, List<String>>` but that is broken and does not work with the current releases.

All of this just to say setting `wrapCollectionElements(true)` at the server/client in the next Infinispan release should be fine.

ryanemerson: @pruivo I thought `RemoteCache<String, List<String>>` would work if the list only contained a single entry?

pruivo: Collection of primitive works with any number of elements. The collection of Pojo doesn't and only works with strictly a single element. In any case, the Infinspan server does not manipulate the byte[] sent by the client except for Indexing/Query.

ryanemerson: and transcoding :slightly_smiling_face:

Anyway, thanks for clarifying. IMO the benefits of fixing this issue far outweigh the consequences of older clients having an incompatibility with a new server version specifically for single entry collections. \\cc @tristantarrant

ryanemerson:
Thanks @pruivo

PR #274, opened 2024-05-15
IPROTO-334 Don't throw NPE when proto.lock contains a message with no…

tristantarrant:
https://issues.redhat.com/browse/IPROTO-334


ryanemerson:
Thanks @tristantarrant

PR #279, opened 2024-06-02
fix: PROTO_SCHEMA generated value with unclosed string literal

israel-ramos:
The issue:
-  The value being generated for constant PROTO_SCHEMA was adding odd quotes in the middle of the string. So the generated value could not be compiled.
image

The cause: 
- The method makeStringLiteral, used to generate the value for PROTO_SCHEMA, was building the string iterating char by char and appending line breaks, slashes, quotes and plus signs in order to transform the string to literal keeping that as multi line.

Suggested fix: 
- The method was refactored to use java builtin multi-line string support with triple quotes (""" ..._multi line string here_... """). So, no need to escape line breaks any more.


tristantarrant:
Nice fix. Thanks

PR #287, opened 2024-06-25
IPROTO-344 NullPointerException in START_GROUP case parser

pruivo:
I fixed a possible NPE in the parser.


fax4ever (approved):
LGMT, thanks @pruivo

PR #291, opened 2024-07-01
IPROTO-343 Allow null values for primitive types with PROTO3 syntax

ryanemerson:
https://issues.redhat.com/browse/IPROTO-343

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/annotations/impl/ProtoMessageTypeMetadata.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/impl/ProtoMessageTypeMetadata.java
@@ -834,7 +834,8 @@ public class ProtoMessageTypeMetadata extends ProtoTypeMetadata {
     * type.
     */
    private Object getDefaultValue(XClass clazz, String fieldName, XClass fieldType, Type protobufType, String value, boolean isRepeated) {
-      if (protoSchemaGenerator.syntax() == ProtoSyntax.PROTO2) {
+      if (protoSchemaGenerator.syntax() == ProtoSyntax.PROTO2 ||
+            (protoSchemaGenerator.allowNullFields() && !fieldType.isPrimitive())) {
          if (value == null || value.isEmpty()) {
             return null;
          }
--- a/core/src/main/java/org/infinispan/protostream/annotations/ProtoSchema.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/ProtoSchema.java
@@ -120,4 +120,17 @@ public @interface ProtoSchema {
     * Specifies the protobuf syntax used by the generated schema. Defaults to {@link ProtoSyntax#PROTO2}
     */
    ProtoSyntax syntax() default ProtoSyntax.PROTO2;
+
+   /**
+    * Specifies whether generated marshallers should allow missing fields of Byte Arrays, Enums, boxed primitives and
+    * Strings to be initialized as null when the {@link ProtoField#defaultValue()} is defined as an empty string.
+    * <p>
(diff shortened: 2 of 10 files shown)


integrationtests/src/test/java/org/infinispan/protostream/integrationtests/processor/marshaller/model/NullsAllowedModel.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class NullsAllowedModel {
+
+   public static final GeneratedSchema SCHEMA = new NullsAllowedSchemaImpl();
+
+   @ProtoField(1)
+   public String string;
+
+   @ProtoField(2)
+   public Integer integer;
```
pruivo, line 18: If I change this to `int`, it fails with
```
Primitive field 'integer' of org.infinispan.protostream.integrationtests.processor.marshaller.model.NullsAllowedModel is not nullable so it should be either marked required or should have a default value, while processing org.infinispan.protostream.integrationtests.processor.marshaller.model.NullsAllowedModel.NullsAllowedSchema
```

core/src/main/java/org/infinispan/protostream/annotations/impl/ProtoMessageTypeMetadata.java
```diff
@@ -834,7 +834,7 @@ protected XClass getJavaTypeFromAnnotation(ProtoField annotation) {
     * type.
     */
    private Object getDefaultValue(XClass clazz, String fieldName, XClass fieldType, Type protobufType, String value, boolean isRepeated) {
-      if (protoSchemaGenerator.syntax() == ProtoSyntax.PROTO2) {
+      if (protoSchemaGenerator.syntax() == ProtoSyntax.PROTO2 || protoSchemaGenerator.allowNullFields()) {
```
pruivo, line 837: this seems to do the trick:
```suggestion
      if (protoSchemaGenerator.syntax() == ProtoSyntax.PROTO2 || (protoSchemaGenerator.allowNullFields() && !fieldType.isPrimitive())) {
```

pruivo (changes requested):
Primitive types aren't working anymore. I suggested a possible fix :crossed_fingers:

pruivo:
ps. would be nice to add a primitive type into the test too :+1:

ryanemerson:
Good spot @pruivo. Fix and test updates pushed.

integrationtests/src/test/java/org/infinispan/protostream/integrationtests/processor/marshaller/model/NullsAllowedModel.java
```diff
@@ -0,0 +1,40 @@
...
+
+   public static final GeneratedSchema SCHEMA = new NullsAllowedSchemaImpl();
+
+   @ProtoField(1)
+   public String string;
+
+   @ProtoField(2)
+   public Integer boxedInt;
+
+   @ProtoField(value = 3, defaultValue = "0")
```
pruivo, line 20: is the `defaultValue="0"` a requirement? Or can we get rid of it? :smile:

ryanemerson: Argh, that was a leftover from some local testing. Removed now.

ryanemerson:
@pruivo Should be good now. I updated the testsuite and `@ProtoSchema` docs to reflect that the null behaviour also applies to Enums and `byte[]`.

pruivo (approved):
LGTM. Tested with Keycloak (https://github.com/pruivo/keycloak/commit/86a0614e3825a475eedea73e7be8374254daadf3)

pruivo:
merged! thanks @ryanemerson !

PR #294, opened 2024-07-02
IPROTO-351 TypeId must be parsed before the compatibility check

pruivo:
https://issues.redhat.com/browse/IPROTO-351

I don't know if this is the best way to do it and the method name is not the best. Suggestions are welcome.


ryanemerson (approved):
LGTM, I would like @tristantarrant to double check as the original author of the feature before we merge though.

tristantarrant:
Nice one, thanks

PR #303, opened 2024-07-15
IPROTO-361 fix default value of Long

ahus1:
This fixes it for me when I build this locally. 

I could use a little help with in writing a test.


ryanemerson:
Thanks @ahus1

PR #333, opened 2024-10-03
IPROTO-265 Avoid byte[] allocation for buffer of OutputStream

wburns:
https://issues.redhat.com/browse/IPROTO-265


wburns:
Note this is a subset of the original commit at https://github.com/infinispan/protostream/commit/16ba83f86219e1ca2228a6bbb4894825bcc16744 so that we avoid the byte[] that is used to buffer writes between the TagWriter and OutputStream as many implementations may not need this. In fact it is much simpler to instead wrap the OutputStream with their own BufferOutputStream if they want.

core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
```diff
@@ -154,7 +154,7 @@ public static ByteBuffer toWrappedByteBuffer(ImmutableSerializationContext ctx,
    }
 
    public static void toWrappedStream(ImmutableSerializationContext ctx, OutputStream out, Object t) throws IOException {
-      toWrappedStream(ctx, out, t, DEFAULT_STREAM_BUFFER_SIZE);
+      WrappedMessage.write(ctx, TagWriterImpl.newInstance(ctx, out), t);
```
pruivo, line 157: shouldn't be `TagWriterImpl.newInstanceNoBuffer`?

pruivo: Ah! I see you changed `newInstance(ImmutableSerializationContext serCtx, OutputStream output)`. 
So `newInstanceNoBuffer` is kind of duplicated/irrelevant at this point!?

wburns: I was not aware of newInstanceNoBuffer! Yeah, no reason newInstance should ever use the buffered version imo. Updated to deprecate newInstanceNoBuffer as well.

pruivo (approved):

pruivo:
merged! thanks @wburns !

PR #336, opened 2024-10-04
IPROTO-394 Optimize TagWriterImpl#writeString for ASCII only Strings

wburns:
https://issues.redhat.com/browse/IPROTO-394

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
@@ -131,13 +131,30 @@ public final class TagWriterImpl implements TagWriter, ProtobufTagMarshaller.Wri
       encoder.writeVarint64(value);
    }
 
+   public static boolean onlyAscii(String ba, int off, int len) {
+      int limit = off + len;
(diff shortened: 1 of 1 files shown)


wburns:
This could be optimized further by streaming the bytes immediately irrespective of contents but would require more work and testing.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -131,13 +131,30 @@ public void writeVarint64(long value) throws IOException {
       encoder.writeVarint64(value);
    }
 
+   public static boolean onlyUTF8(String ba, int off, int len) {
```
tristantarrant, line 134: Shouldn't this method be called `isAsciiOnly ()` ?

wburns: Yeah, I didn't push my update ;)

wburns:
Changed to draft as I am seeing an adverse performance impact in my test.. despite it supposedly only using ASCII keys and values. Have to investigate further.

PR #358, opened 2024-10-31
build from release commit

hboutemy:
currently, Git commit reference in release binaries output is one commit behind the Git tag, given the order of operation in the workflow

Git commit has to be done before building: perhaps it won't be pushed if build fails, but at least build references the right commit hash that will get tagged

issue found while trying to rebuild for Reproducible Builds: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/infinispan/protostream/README.md
and more precisely analysis differences https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/infinispan/protostream/protostream-aggregator-5.0.7.Final.diffoscope


ryanemerson:
Thanks @hboutemy

PR #401, opened 2025-01-10
[#394] Don't use generics when marshalling collections

### LINKED ISSUE #394: Demarshalling doesn't work for collections of generic objects
I've found a problem with the protostream marshallers. The following class generates a compiler error in its marshaller:
```java
public class DTO {
  // other fields

  private Set<Pair<String, String>> restrictions;

  // other getters

  @ProtoField(number = 12, collectionImplementation = HashSet.class)
  public Set<Pair<String, String>> getRestrictions() {
     return restrictions;
  }
}
```
Marshaller method (Version 5.x):
```java
public void write(WriteContext $1,  DTO $2) {
      // writing of other fields is ok
      {
         final java.util.Collection<Pair> __c$13 = o.getRestrictions(); /* <= this does not work because Pair is not parameterized 
                                                           with <String, String> while the method o.getRestrictions() returns a Set<Pair<String, String>> */
         if (__c$13 != null) 
            for (java.util.Iterator<Pair> it = __c$13.iterator(); it.hasNext(); ) {
               final Pair __v$13 = it.next();
               {
                  if (__md$13 == null) __md$13 = ((org.infinispan.protostream.impl.SerializationContextImpl) $1.getSerializationContext()).getMarshallerDelegate(Pair.class);
                  writeNestedMessage(__md$13, $out, 13, __v$13);
               }
            }
      }
}
```
This only seems to pose a problem since Protostream 5.x because in version 4.x it worked because the generated `write`-method did not use a parameterized collection. Instead it contained a loop over all elements of the collection like this:
Marshaller method (Version 4.x):
```java
{
            final java.util.Collection __c$13 = o.getRestrictions();
            if (__c$13 != null) for (java.util.Iterator it = __c$13.iterator(); it.hasNext(); ) {
                final Pair __v$13 = (Pair) it.next();
                {
                    if (__md$13 == null)
                        __md$13 = ((org.infinispan.protostream.impl.SerializationContextImpl) $1.getSerializationContext()).getMarshallerDelegate(Pair.class);
                    writeNestedMessage(__md$13, $1, 13, __v$13);
                }
            }
        }
```

tristantarrant:
Closes #394

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/annotations/impl/AbstractMarshallerCodeGenerator.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/impl/AbstractMarshallerCodeGenerator.java
@@ -47,6 +47,8 @@ public abstract class AbstractMarshallerCodeGenerator {
 
    private final String protobufSchemaPackage;
 
+   private final boolean useGenerics = false;
+
    protected AbstractMarshallerCodeGenerator(XTypeFactory typeFactory, String protobufSchemaPackage) {
       this.typeFactory = typeFactory;
       this.protobufSchemaPackage = protobufSchemaPackage;
@@ -719,7 +721,11 @@ public abstract class AbstractMarshallerCodeGenerator {
                         mapFieldMetadata.getValue().getJavaTypeName()
                   );
                } else {
-                  iw.printf("java.util.Collection<%s>", fieldMetadata.getJavaTypeName());
+                  if (useGenerics) {
+                     iw.printf("java.util.Collection<%s>", fieldMetadata.getJavaTypeName());
+                  } else {
+                     iw.printf("java.util.Collection");
+                  }
                }
             } else {
                iw.print(fieldMetadata.getJavaTypeName());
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/infinispan/protostream/annotations/impl/AbstractMarshallerCodeGenerator.java
```diff
@@ -719,7 +721,11 @@ protected void generateWriteMethodBody(IndentWriter iw, ProtoMessageTypeMetadata
                         mapFieldMetadata.getValue().getJavaTypeName()
                   );
                } else {
-                  iw.printf("java.util.Collection<%s>", fieldMetadata.getJavaTypeName());
+                  if (useGenerics) {
```
ryanemerson, line 724: This is always false.

tristantarrant: I know, I wanted to keep the code around for when I do it properly

ryanemerson:
Thanks @tristantarrant, please backport to `5.0.x`

PR #415, opened 2025-02-05
Performance improvements

### LINKED ISSUE #413: Optimize TagWriterImpl#writeString for ASCII only Strings
Currently TagWriterImpl always calls getBytes passing the UTF8 Charset. This allocates a byte[] to store these values. We should be able to instead detect if the String is only ASCII and if so write directly to the encoder without allocating the byte[].

### LINKED ISSUE #414: Replace ByteArrayOutputStreamEx with non-synchronized implementation
`ByteArrayOutputStreamEx` is used throughout Protostream as the default OutputStream. While it's possible for users to provide their own `OutputStream` implementation via `ProtobufUtil`, it's not possible for them to configure the one used in the internal/generated marshaller classes.

Benchmarking shows that `ByteArrayOutputStreamEx` is considerably slower than non-thread safe `OutputStream` implementations due to the use of `synchronized` methods. As Protostream does not take advantage of multi-threading, we should replace `ByteArrayOutputStreamEx` with a non-thread safe implementation that does not utilise `synchronized`.

ryanemerson:
Closes #413
Closes #414


tristantarrant:
For reference it would be nice to have some benchmarks between old and new implementations

core/src/main/java/org/infinispan/protostream/LazyByteArrayOutputStream.java
```diff
@@ -0,0 +1,27 @@
+package org.infinispan.protostream;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+public interface LazyByteArrayOutputStream {
```
pruivo, line 6: nitpicking but can we get a new name? I have no idea what lazy means in this context. 

My not-so-great suggestion is `ProtoOutputStream`

Javadoc would be nice since this is a public API.

core/src/main/java/org/infinispan/protostream/impl/ByteArrayOutputStreamEx.java
```diff
@@ -8,7 +8,9 @@
  *
  * @author anistor@redhat.com
  * @since 4.0
+ * @deprecated Use {@link LazyByteArrayOutputStream} instead.
  */
+@Deprecated(forRemoval = true)
```
pruivo, line 13: seeing this in the `impl` package, I'm tempted to say "remove it". 
Are we planning to backport this PR/optimizations for ProtoStream 5.x? We can deprecate it there and remove it here.

core/src/main/java/org/infinispan/protostream/impl/LazyByteArrayOutputStream.java
```diff
@@ -0,0 +1,83 @@
...
+   }
+
+   byte[] buf;
+   int pos = 0;
+
+   @Override
+   public void write(int b) {
+      ensureCapacity(pos + 1);
+      buf[pos] = (byte) b;
+      pos += 1;
```
pruivo, line 29: you're computing the new position twice in all the methods `write(byte)`and `write(byte[])`. I'm not sure if it will make any difference:
```java
      var newpos = pos + 1;
      ensureCapacity(newpos);
      buf[pos] = (byte) b;
      pos = newpos;
```

core/src/main/java/org/infinispan/protostream/impl/LazyByteArrayOutputStream.java
```diff
@@ -0,0 +1,83 @@
...
+      ensureCapacity(pos + len);
+      System.arraycopy(b, 0, buf, pos, len);
+      pos += len;
+   }
+
+   @Override
+   public void write(byte[] b, int off, int len) {
+      if (len == 0)
+         return;
+
```
pruivo, line 44: safety first? `System.arraycopy` will probably do the check too... so, let me know what you think. 
```java
Objects.checkFromIndexSize(off, len, b.length);
```

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -794,6 +819,51 @@ void writeBytes(ByteBuffer value) throws IOException {
...
+         }
+
+         int strlen = s.length();
+         int tag = WireType.makeTag(number, WireType.WIRETYPE_LENGTH_DELIMITED);
+         int varIntTagLen = varIntSize(tag);
+         int varIntBytesLen = varIntSize(strlen);
+         int startPos = os.getPosition();
+
+         // First optimize for 1 - 127 case
+         os.ensureCapacity(startPos + varIntTagLen + varIntBytesLen + strlen);
```
pruivo, line 836: why not assume the worst case of 5 bytes per varInt to avoid computing it? With the array growing double in size, I don't expect it to create growth more often.

core/src/test/java/org/infinispan/protostream/impl/TagWriterImplTest.java
```diff
@@ -22,7 +23,57 @@ public class TagWriterImplTest {
...
+
+   @Test
+   public void testUTF8Compatibility() throws IOException {
+      SerializationContext ctx = ProtobufUtil.newSerializationContext();
+      testUTF8String(ctx, utf8Str(10, false));
+      testUTF8String(ctx, utf8Str(100, false));
+      testUTF8String(ctx, utf8Str(1000, false));
```
pruivo, line 34: I would add an extra test with random length
```java
testUTF8String(ctx, utf8Str(ThreadLocalRandom.current().nextInt(100000000), false));
```
ps log the value just in case it fails we know the length to use :smile:

pruivo (changes requested):
I don't think you are going to like my suggestion. 

I suggest to create a new `org.infinispan.protostream.impl.TagWriterImpl.Encoder` with the "LazyByteArrayOutputStream". I think all the methods would benefit from it and avoid doing `instanceOf` in the `writeUTF8Field` method.

For example, the method is invoking `ensureCapacity` 8 times while we can do better
```java
      void writeFixed64(long value) throws IOException {
         out.write((byte) (value & 0xFF));
         out.write((byte) ((value >> 8) & 0xFF));
         out.write((byte) ((value >> 16) & 0xFF));
         out.write((byte) ((value >> 24) & 0xFF));
         out.write((byte) ((int) (value >> 32) & 0xFF));
         out.write((byte) ((int) (value >> 40) & 0xFF));
         out.write((byte) ((int) (value >> 48) & 0xFF));
         out.write((byte) ((int) (value >> 56) & 0xFF));
      }
``` 

Even the default impl of the `Encoder` class can benefit
```java
      void writeFixed64Field(int fieldNumber, long value) throws IOException {
         writeVarint32(WireType.makeTag(fieldNumber, WireType.WIRETYPE_FIXED64));
         writeFixed64(value);
      }
```
^ for example, ensure 13 bytes capacity, write directly to the buffer, and update the position at the end.

ryanemerson:
> For reference it would be nice to have some benchmarks between old and new implementations

You're right! Below are the figures I posted on Zulip the other day which tests UTF8 marshalling in isolation:

```
Benchmark                         (initialArraySize)  (initialPosition)  (stringLength)      (type)  (useMultiByte)   Mode  Cnt         Score         Error  Units
UtfBenchmark.testUtfWrite                         64                  0               8    proto-ex           false  thrpt    5  17293297.695 ±  775157.948  ops/s
UtfBenchmark.testUtfWrite                         64                  0               8  proto-lazy           false  thrpt    5  28974585.735 ±  144449.312  ops/s
```

`proto-ex` is the old code, `proto-lazy` new.

I will re-run the numbers once I have looked into Pedro's suggestions.

core/src/main/java/org/infinispan/protostream/impl/ByteArrayOutputStreamEx.java
ryanemerson: > Are we planning to backport this PR/optimizations for ProtoStream 5.x? We can deprecate it there and remove it here.

I'm not sure. My gut feeling would be that it's not worth backporting, given that the main driver is the migration to protostream and the fact we only want new  features on `main`.

+1 to deprecating in `5.0.x`.

core/src/main/java/org/infinispan/protostream/LazyByteArrayOutputStream.java
ryanemerson: I've gone with `RawByteArrayOutputStream` and `RawByteArrayOutputStreamImpl` for now.

core/src/main/java/org/infinispan/protostream/LazyByteArrayOutputStream.java
```diff
@@ -0,0 +1,27 @@
...
+   int getPosition();
+   void setPosition(int position);
+   void ensureCapacity(int size);
+   byte[] getRawBuffer();
```
wburns, line 10: Is there any reason we aren't just doing `write(int pos, byte val)` instead? This would allow for other implementations that aren't byte[] backed.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -794,6 +819,51 @@ void writeBytes(ByteBuffer value) throws IOException {
...
+         int c;
+         int i;
+         for (i = 0; i < strlen; i++) {
+            c = s.charAt(i);
+            if (c > 127) break;
+
+            buf[localPos++] = (byte) c;
+         }
+
+         os.setPosition(localPos);
```
wburns, line 853: I believe this only has to be in the if statement below.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -794,6 +819,51 @@ void writeBytes(ByteBuffer value) throws IOException {
...
+         // All single byte characters
+         if (i == strlen)
+            return;
+
+         byte[] utf8buffer = s.getBytes(StandardCharsets.UTF_8);
+         varIntBytesLen = varIntSize(utf8buffer.length);
+         os.ensureCapacity(startPos + varIntTagLen + varIntBytesLen + utf8buffer.length);
+         buf = os.getRawBuffer();
+         writeVarInt32Direct(utf8buffer.length, startPos + varIntTagLen, buf);
+         System.arraycopy(utf8buffer, 0, buf, startPos + varIntTagLen + varIntBytesLen, utf8buffer.length);
```
wburns, line 863: Shouldn't this be `out.writeBytes(uft8Buffer, 0, buf)`? Then the setPosition call wouldn't be required either.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
ryanemerson: On my local benchmarks calculating the size resulted in better performance than pre-allocating.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
ryanemerson: I think it's similar, as calling `os.write()` will result in `ensureCapacity` being called internally again, which we know is sufficient as we have already expanded the buf (if required).

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
ryanemerson: We need it in either case. If it's a multi-byte string, then we call `#ensureCapacity` which utilises the position to copy all existing bytes to the array created with the new capacity.

core/src/main/java/org/infinispan/protostream/LazyByteArrayOutputStream.java
ryanemerson: The original purpose of the interface was to allow different implementations of `byte[]` backed streams, so exposing `byte[]` directly removes the need for lots of additional `#write` methods calls.

I guess your suggestion would allow the interface to be used by other indexed `OutputStream` implementations though, e.g. one that is ByteBuf based. Would this be beneficial for the client?

wburns: Yes, this is what I am thinking. I think this should benefit both the client and possibly even the server with some more changes I am thinking about.

core/src/main/java/org/infinispan/protostream/impl/RawByteArrayOutputStreamImpl.java
```diff
@@ -0,0 +1,93 @@
+package org.infinispan.protostream.impl;
+
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+import net.jcip.annotations.NotThreadSafe;
+
+@NotThreadSafe
+public class RawByteArrayOutputStreamImpl extends OutputStream implements org.infinispan.protostream.RawByteArrayOutputStream {
```
pruivo, line 9: nitpicking
```suggestion
public class RawByteArrayOutputStreamImpl extends OutputStream implements RawByteArrayOutputStream {
```

core/src/main/java/org/infinispan/protostream/impl/RawByteArrayOutputStreamImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+   static final int MIN_SIZE = 32;
+
+   static final int DEFAULT_DOUBLING_SIZE = 4 * 1024 * 1024; // 4MB
+
+   byte[] buf;
+   int pos = 0;
+
+   public RawByteArrayOutputStreamImpl() {
+   }
```
pruivo, line 19: shouldn't create a new a byte[] with default size (`MIN_SIZE`)? We could skip the null checks in `ensureCapacity`and `getByteBuffer`.

core/src/main/java/org/infinispan/protostream/impl/RawByteArrayOutputStreamImpl.java
```diff
@@ -0,0 +1,93 @@
...
+      pos = newpos;
+   }
+
+   @Override
+   public void write(byte[] b) {
+      final int len = b.length;
+      ensureCapacity(pos + len);
+      System.arraycopy(b, 0, buf, pos, len);
+      pos += len;
+   }
```
pruivo, line 41: the default implementation from `OutputStream` would do fine.

core/src/main/java/org/infinispan/protostream/impl/RawByteArrayOutputStreamImpl.java
pruivo: Why extend `OutputStream`?

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -915,4 +931,295 @@ void flush() throws IOException {
...
+      void writeFixed32Field(int fieldNumber, int value) {
+         int pos = out.getPosition();
+         int tag = WireType.makeTag(fieldNumber, WireType.WIRETYPE_FIXED32);
+         int tagLen = varIntBytes(tag);
+         out.ensureCapacity(pos + tagLen + FIXED_32_SIZE);
+
+         byte[] buf = out.getRawBuffer();
+         pos = writeVarInt32Direct(buf, pos, tag);
+         pos = writeFixed32Direct(buf, pos, value);
+         out.setPosition(pos);      }
```
pruivo, line 984:
```suggestion
         out.setPosition(pos);      
      }
```

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -915,4 +931,295 @@ void flush() throws IOException {
...
+         if (value.hasArray()) {
+            int length = value.remaining();
+            writeBytes(fieldNumber, value.array(), value.arrayOffset() + value.position(), length);
+            value.position(value.position() + length);
+         } else {
+            int bbPos = value.position();
+            int bbLen = bbPos + value.remaining();
+
+            int tag = WireType.makeTag(fieldNumber, WireType.WIRETYPE_LENGTH_DELIMITED);
+            out.ensureCapacity(pos + MAX_VARINT_SIZE + MAX_VARINT_SIZE + bbLen);
```
pruivo, line 1038: I'm wondering why in some places you use `varIntBytes` and in another `MAX_VARINT_SIZE`.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -915,4 +931,295 @@ void flush() throws IOException {
...
+               value >>>= 7;
+            }
+         }
+         return i;
+      }
+
+      private int varIntBytes(long value) {
+         int i = 1;
+         while (true) {
+            if ((value & 0xFFFFFF80) == 0) {
```
pruivo, line 1216:
```suggestion
            if ((value & 0xFFFFFF80L) == 0) {
```

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -915,4 +931,295 @@ void flush() throws IOException {
...
+         varIntBytesLen = varIntBytes(utf8buffer.length);
+         out.ensureCapacity(startPos + varIntTagLen + varIntBytesLen + utf8buffer.length);
+
+         buf = out.getRawBuffer();
+         writeVarInt32Direct(buf, startPos + varIntTagLen, utf8buffer.length);
+         System.arraycopy(utf8buffer, 0, buf, startPos + varIntTagLen + varIntBytesLen, utf8buffer.length);
+         out.setPosition(startPos + varIntTagLen + varIntBytesLen + utf8buffer.length);
+      }
+
+      private int writeFixed32Direct(byte[] buf, int i, int value) {
```
pruivo, line 1167: all the write methods from here down can be static

core/src/main/java/org/infinispan/protostream/impl/UnknownFieldSetImpl.java
```diff
@@ -198,7 +199,7 @@ public boolean hasTag(int tag) {
 
    @Override
    public void writeExternal(ObjectOutput out) throws IOException {
-      ByteArrayOutputStreamEx baos = new ByteArrayOutputStreamEx();
+      RawByteArrayOutputStream baos = new RawByteArrayOutputStreamImpl();
       TagWriter output = TagWriterImpl.newInstance(null, baos);
       writeTo(output);
       output.flush();
```
pruivo, line 205: Lines 206-210 can be replaced with
```java
      out.writeInt(baos.getPosition());
      out.write(baos.getRawBuffer(), 0, baos.getPosition());
```

pruivo (changes requested):

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
ryanemerson: Something I missed with all the refactoring :upside_down_face:

core/src/main/java/org/infinispan/protostream/impl/UnknownFieldSetImpl.java
ryanemerson: Now that we're not exposing `getRawBuffer` this is no longer possible. Does it make more sense to wrap the provided `ObjectOutput` in an Adapter and pass this to `TagWriterImpl#newInstance`?

core/src/main/java/org/infinispan/protostream/impl/RawByteArrayOutputStreamImpl.java
ryanemerson: During my benchmarks it seem liked initializing the array was leading to worse performance. I believe this is because we're creating a lot of stream instances and in many cases the initial capacity required exceeded `MIN_SIZE`.

core/src/main/java/org/infinispan/protostream/impl/RawByteArrayOutputStreamImpl.java
ryanemerson: It's not strictly necessary, but it makes changing/benchmarking implementations easier when `OutputStream` is a common base class.

core/src/main/java/org/infinispan/protostream/impl/UnknownFieldSetImpl.java
pruivo: `getRawBuffer` is still a public method. 
At the moment, I don't see how the adapter would work as you don't know the size before hand.

ryanemerson:
I have updated the interface to not expose the raw `byte[]` as suggested by @wburns so that we can utilise the stream for ByteBuf based solutions as well. Unfortunately it seems like there's a perf hit when compared with the values I posted yesterday:

```

Benchmark                         (initialArraySize)  (initialPosition)  (stringLength)      (type)  (useMultiByte)   Mode  Cnt         Score        Error  Units
UtfBenchmark.testUtfWrite                         32                  0               8  proto-lazy            true  thrpt    5  18189985.935 ± 708413.260  ops/s
UtfBenchmark.testUtfWrite                         32                  0               8  proto-lazy           false  thrpt    5  34802151.363 ± 543860.030  ops/s
UtfBenchmark.testUtfWrite                         32                  0               8    proto-ex            true  thrpt    5  11461152.933 ± 569242.019  ops/s
UtfBenchmark.testUtfWrite                         32                  0               8    proto-ex           false  thrpt    5  14165941.018 ±  54503.091  ops/s
```

core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
```diff
@@ -70,18 +69,22 @@ private static <A> void write(ImmutableSerializationContext ctx, TagWriterImpl o
       out.flush();
    }
 
+   public static void writeTo(ImmutableSerializationContext ctx, RandomAccessOutputStream out, Object t) throws IOException {
+      write(ctx, TagWriterImpl.newInstance(ctx, out), t);
+   }
+
    public static void writeTo(ImmutableSerializationContext ctx, OutputStream out, Object t) throws IOException {
       write(ctx, TagWriterImpl.newInstance(ctx, out), t);
```
pruivo, line 77: nitpicking, you probably want to do `out instanceof RandomAccessOutputStream` as you did in `toWrappedStream` method.

core/src/main/java/org/infinispan/protostream/impl/RandomAccessOutputStreamImpl.java
```diff
@@ -0,0 +1,103 @@
...
+      if (buf == null) {
+         buf = new byte[Math.max(MIN_SIZE, capacity)];
+      } else if (capacity > buf.length) {
+         byte[] newbuf = new byte[getNewBufferSize(buf.length, capacity)];
+         System.arraycopy(buf, 0, newbuf, 0, pos);
+         buf = newbuf;
+      }
+   }
+
+   private int getNewBufferSize(int curSize, int minNewSize) {
```
pruivo, line 75: nitpicking: can be static.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -915,4 +931,276 @@ void flush() throws IOException {
...
+         while (true) {
+            if ((value & 0xFFFFFF80L) == 0) {
+               return i;
+            } else {
+               i++;
+               value >>>= 7;
+            }
+         }
+      }
+   }
```
pruivo, line 1205: nitpicking, can be static and without the if branch

```suggestion
      private static int varIntBytes(long value) {
         int i = 1;
         while ((value & 0xFFFFFFFFFFFFFF80L) != 0) {
            ++i;
            value >>>= 7;
         }
         return i;
      }
```

core/src/main/java/org/infinispan/protostream/impl/UnknownFieldSetImpl.java
```diff
@@ -198,15 +198,15 @@ public boolean hasTag(int tag) {
...
-      ByteBuffer buffer = baos.getByteBuffer();
-      int off = buffer.arrayOffset();
-      int len = buffer.limit() - off;
-      out.writeInt(len);
-      out.write(buffer.array(), off, len);
+
+      int pos = baos.getPosition();
+      out.writeInt(pos);
+      for (int i = 0; i < pos; i++)
+         out.write(baos.get(i));
```
pruivo, line 209: I took a closer look, and I suggest creating a method `void copyTo(DataOutput output) throws IOException;` in `RandomAccessOutputStream`

Writing byte by byte does not seem very efficient, mainly it has to check the capacity of the `DataOutput` on every invocation.

pruivo (changes requested):
I suggested an "improvement" for `UnknownFieldSetImpl`. The remaining comments are optional.

pruivo (approved):
LGTM! Nice work!

core/src/main/java/org/infinispan/protostream/LazyByteArrayOutputStream.java
ryanemerson: Interface and implementation updated to no longer expose a `byte[]`. Let me know if we're missing anything you think we'll need for your future ideas.

wburns (approved):
LGTM!

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -915,4 +931,273 @@ void flush() throws IOException {
...
+               out.write(i++, (byte) ((int) value & 0x7F | 0x80));
+               value >>>= 7;
+            }
+         }
+         return i;
+      }
+
+      private static int varIntBytes(long value) {
+         int i = 1;
+         while ((value & 0xFFFFFFFFFFFFFF80L) != 0) {
```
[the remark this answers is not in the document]
pruivo: 8 ifs, this has to support long too.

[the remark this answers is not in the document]
pruivo: how so? the max var int size is 10 bytes. How can you reduce that to 4 ifs?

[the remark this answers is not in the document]
wburns: Wrote a quick JMH benchmark and for a single byte they have the same perf. Only once I increased it to something larger was there much of a difference and it still only a ns at worst (.2ns for the ifs and 1.13ns for the loop for 5 bytes). Note the ifs were the same perf no matter the size, guessing due to branch prediction. Either way as I mentioned the difference is so minor it probably isn't worth it. Especially as the code here works for int and long and is much more concise.

PR #426, opened 2025-02-26
[#425] Git commit message template and validation

### LINKED ISSUE #425: Standard git commit message and validation hook
### Description

To ensure homogeneity in our commit message, we should introduce a validation hook
Implementation ideas

This is my proposed format for the message:

`[#01234] Title, summary, imperative, start upper, max 50`

* Item 1 wrap at 72 characters
  wrapped text
* Item 2

There are two places where we can run this validation: either locally, as a commit-msg git hook or server-side, as a pre-receive git hook.
The former would let us skip the validation for exceptional situations via the git commit --no-verify switch, and is my preference

The following regex should work for the above format

`^\[#[0-9]+\]\s[A-Z].*\n(\n(\*\s.*\n( .+\n)*)+)?$`

although it doesn't validate line length

tristantarrant:
Closes #425


ryanemerson:
Thanks @tristantarrant

PR #428, opened 2025-02-26
Optimize nested writers during marshalling.

### LINKED ISSUE #427: Refactor nested messages to allow writer/encoder handle logic
Today when we have a nested message we always create a new TagWriter without knowledge of what the current writer is. This usually entails us creating a brand new TagWriter with a byte[] backed stream. This means each nested message incurs additional byte[] allocation and the cost to copy that byte[] back into the original writer. This can happen multiple layers down as well.

We should be able to refactor the code to move the logic of handling the new tag writer to the existing writer so it can allocate a more efficient writer as necessary or even reuse itself. Reuse though needs to be careful about ensuring nested depth is adhered to though.

wburns:
Fixes #427

This PR introduces a new method to TagWriter and Encoder to allow those classes control the TagWriter and Encoder that are created whenever a sub message is found. It removes all manual creations of TagWriters in those places and replaces with calls to the new method.

The first commit does that refactoring and by default has all Writer/Encoders use OutputStreamRandomAccessEncoder as was done prior. The big difference is the new encoders will cache previously used sub encoders so they will not have to recreate them for multiple nested messages at the same depth.

The second commit takes this one step further by introducing a move method to the RandomAccessOutputStream by allowing it to move bytes within the same stream. This allows us to not have to create any additional byte[] and instead it reuses the same one from the parent.

The third commit is just one to remove some unneeded computations for var ints where it was finding the exact size to ensure the byte[] was large enough. In that case there is no real reason to incur the true cost and just assume 5 for the worst case. In many cases this will not cause a resize and the cases when it does it most likely would have had to resize for the next value anyways.

The fourth commit takes the second commit a step further by actually reusing the same Encoder instance for all nested messages. Instead the previous sizes are stored in an int[] and when close is invoked it takes the last size value to write to it, further reducing allocation costs.

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/annotations/impl/AbstractMarshallerCodeGenerator.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/impl/AbstractMarshallerCodeGenerator.java
@@ -23,7 +23,6 @@ import org.infinispan.protostream.annotations.impl.types.XTypeFactory;
 import org.infinispan.protostream.descriptors.JavaType;
 import org.infinispan.protostream.descriptors.Type;
 import org.infinispan.protostream.descriptors.WireType;
-import org.infinispan.protostream.impl.TagWriterImpl;
 
 /**
  * @author anistor@readhat.com
@@ -712,7 +711,7 @@ public abstract class AbstractMarshallerCodeGenerator {
          getUnknownFieldSetFieldStatement = "o.getUnknownFieldSet()";
       }
       if (!messageTypeMetadata.getFields().isEmpty() || getUnknownFieldSetFieldStatement != null) {
-         iw.printf("%s $out = (%s) $1.getWriter();\n", TagWriterImpl.class.getName(), TagWriterImpl.class.getName());
+         iw.print("var $out = $1.getWriter();\n");
          iw.printf("final %s o = (%s) $2;\n", messageTypeMetadata.getJavaClassName(), messageTypeMetadata.getJavaClassName());
          for (ProtoFieldMetadata fieldMetadata : messageTypeMetadata.getFields().values()) {
             iw.println("{");
@@ -778,14 +777,12 @@ public abstract class AbstractMarshallerCodeGenerator {
                         mapFieldMetadata.getValue().getJavaTypeName(),
                         v
                   );
-                  iw.printf("try (NestedWriter $n = new NestedWriter($1, %d)) {\n", fieldMetadata.getNumber());
+                  iw.printf("try (var $n = $out.subWriter(%d, true)) {\n", fieldMetadata.getNumber());
                   iw.inc();
-                  iw.println("$out = $n.getWriter();");
-                  writeFieldValue(mapFieldMetadata.getKey(), iw, v + ".getKey()", "$out");
-                  writeMapFieldValue(mapFieldMetadata, mapFieldMetadata.getValue(), iw, v + ".getValue()", "$out");
+                  writeFieldValue(mapFieldMetadata.getKey(), iw, v + ".getKey()", "$n");
+                  writeMapFieldValue(mapFieldMetadata, mapFieldMetadata.getValue(), iw, v + ".getValue()", "$n");
                   iw.dec();
                   iw.println("}");
-                  iw.printf("$out = (%s) $1.getWriter();\n", TagWriterImpl.class.getName());
                } else {
                   if (useGenerics) {
                      iw.printf("for (java.util.Iterator<%s> it = %s.iterator(); it.hasNext(); ) {\n", fieldMetadata.getJavaTypeName(), f);
@@ -865,7 +862,7 @@ public abstract class AbstractMarshallerCodeGenerator {
             iw.println("{");
             iw.inc();
             String mdField = initMarshallerDelegateField(iw, fieldMetadata);
-            iw.printf("writeNestedMessage(%s, %s, %d, %s);\n", mdField, out, fieldMetadata.getNumber(), v);
+            iw.printf("writeNestedMessage(%s, (WriteContext) %s, %d, %s);\n", mdField, out, fieldMetadata.getNumber(), v);
             iw.dec();
             iw.println("}");
             break;
@@ -892,7 +889,7 @@ public abstract class AbstractMarshallerCodeGenerator {
             iw.inc();
             fieldName = makeMarshallerDelegateFieldName(mapMetadata);
             mdField = initMarshallerDelegateField(iw, fieldName, fieldMetadata);
-            iw.printf("writeNestedMessage(%s, %s, %d, %s);\n", mdField, out, fieldMetadata.getNumber(), v);
+            iw.printf("writeNestedMessage(%s, (WriteContext) %s, %d, %s);\n", mdField, out, fieldMetadata.getNumber(), v);
             iw.dec();
             iw.println("}");
             break;
--- a/core/src/main/java/org/infinispan/protostream/annotations/impl/GeneratedMarshallerBase.java
+++ b/core/src/main/java/org/infinispan/protostream/annotations/impl/GeneratedMarshallerBase.java
@@ -3,11 +3,9 @@ package org.infinispan.protostream.annotations.impl;
 import java.io.IOException;
 
 import org.infinispan.protostream.ProtobufTagMarshaller;
-import org.infinispan.protostream.RandomAccessOutputStream;
+import org.infinispan.protostream.TagWriter;
 import org.infinispan.protostream.impl.BaseMarshallerDelegate;
 import org.infinispan.protostream.impl.Log;
-import org.infinispan.protostream.impl.RandomAccessOutputStreamImpl;
-import org.infinispan.protostream.impl.TagWriterImpl;
 
 /**
  * Base class for generated message marshallers. Provides some handy helper methods.
@@ -46,30 +44,8 @@ public class GeneratedMarshallerBase {
       if (ctx.depth() >= maxNestedMessageDepth) {
          throw log.maxNestedMessageDepth(maxNestedMessageDepth, message.getClass());
       }
-      try (NestedWriter nested = new NestedWriter(ctx, fieldNumber)) {
-         writeMessage(marshallerDelegate, nested.writer, message);
-      }
-   }
-
-   public static class NestedWriter implements AutoCloseable {
-      final RandomAccessOutputStream baos;
-      final TagWriterImpl writer;
-      private final ProtobufTagMarshaller.WriteContext ctx;
-      private final int fieldNumber;
-
-      public NestedWriter(ProtobufTagMarshaller.WriteContext ctx, int fieldNumber) {
-         this.ctx = ctx;
-         this.fieldNumber = fieldNumber;
-         this.baos = new RandomAccessOutputStreamImpl();
-         this.writer = TagWriterImpl.newNestedInstance(ctx, baos);
-      }
-
-      public TagWriterImpl getWriter() {
-         return writer;
-      }
-
-      public void close() throws IOException {
-         ctx.getWriter().writeBytes(fieldNumber, baos.getByteBuffer());
+      try (TagWriter nestedWriter = ctx.getWriter().subWriter(fieldNumber, true)) {
+         marshallerDelegate.marshall((ProtobufTagMarshaller.WriteContext) nestedWriter, null, message);
       }
    }
 }
--- a/core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
@@ -1134,8 +1134,6 @@ public final class TagWriterImpl implements TagWriter, ProtobufTagMarshaller.Wri
    private static class OutputStreamRandomAccessEncoder extends Encoder {
 
       final RandomAccessOutputStream out;
-      int fieldNumber = -1;
-      OutputStreamRandomAccessEncoder childEncoder = null;
 
       public OutputStreamRandomAccessEncoder(RandomAccessOutputStream out) {
          this.out = out;
@@ -1363,7 +1361,7 @@ public final class TagWriterImpl implements TagWriter, ProtobufTagMarshaller.Wri
          return i;
       }
 
-      private int writeVarInt32Direct(int i, int value) throws IOException {
+      int writeVarInt32Direct(int i, int value) throws IOException {
          while (true) {
             if ((value & 0xFFFFFF80) == 0) {
                out.write(i++, (byte) value);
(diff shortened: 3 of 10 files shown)


wburns:
Running the most up to date benchmark at https://github.com/infinispan/infinispan-benchmarks/tree/main/protostream I got the following numbers compared with main and the new PR. In this case it is ns/op so a lower value is better. In this case the new code is anywhere from 16% to nearly 3x faster.

MAIN

Benchmark | (byteArrayOrStream) | (userByteArraySize) | Mode | Cnt | Score |   | Error | Units
-- | -- | -- | -- | -- | -- | -- | -- | --
ProtostreamBenchmark.testMarshallAddress | TRUE | N/A | avgt | 10 | 82.104 | ± | 0.346 | ns/op
ProtostreamBenchmark.testMarshallAddress | FALSE | N/A | avgt | 10 | 66.607 | ± | 0.423 | ns/op
ProtostreamBenchmark.testMarshallIracMetadata | TRUE | N/A | avgt | 10 | 214.432 | ± | 1.035 | ns/op
ProtostreamBenchmark.testMarshallIracMetadata | FALSE | N/A | avgt | 10 | 192.916 | ± | 0.73 | ns/op
ProtostreamBenchmark.testMarshallUser | TRUE | 10 | avgt | 10 | 479.489 | ± | 1.669 | ns/op
ProtostreamBenchmark.testMarshallUser | TRUE | 8096 | avgt | 10 | 1881.338 | ± | 7.82 | ns/op
ProtostreamBenchmark.testMarshallUser | FALSE | 10 | avgt | 10 | 455.762 | ± | 2.609 | ns/op
ProtostreamBenchmark.testMarshallUser | FALSE | 8096 | avgt | 10 | 1028.067 | ± | 1.778 | ns/op

NEW

Benchmark | (byteArrayOrStream) | (userByteArraySize) | Mode | Cnt | Score |   | Error | Units
-- | -- | -- | -- | -- | -- | -- | -- | --
ProtostreamBenchmark.testMarshallAddress | TRUE | N/A | avgt | 10 | 69.325 | ± | 0.177 | ns/op
ProtostreamBenchmark.testMarshallAddress | FALSE | N/A | avgt | 10 | 54.123 | ± | 0.156 | ns/op
ProtostreamBenchmark.testMarshallIracMetadata | TRUE | N/A | avgt | 10 | 99.783 | ± | 0.318 | ns/op
ProtostreamBenchmark.testMarshallIracMetadata | FALSE | N/A | avgt | 10 | 97.352 | ± | 1.037 | ns/op
ProtostreamBenchmark.testMarshallUser | TRUE | 10 | avgt | 10 | 276.391 | ± | 0.816 | ns/op
ProtostreamBenchmark.testMarshallUser | TRUE | 8096 | avgt | 10 | 1473.529 | ± | 2.012 | ns/op
ProtostreamBenchmark.testMarshallUser | FALSE | 10 | avgt | 10 | 242.012 | ± | 0.663 | ns/op
ProtostreamBenchmark.testMarshallUser | FALSE | 8096 | avgt | 10 | 352.997 | ± | 1.907 | ns/op

The tests done stress specific things

Address - most basic (has no nested messages other than the wrapped message around it)
IracMetadata - has many nested and sub nested messages
User - has some nested messages and a very large byte[] in one case

Also to note the byteArrayOrStream method is if the final destination is to write the message to a byte[] and then copy it into another RandomByteArrayOutpuStream or if the message is directly copied into a RandomByteArrayOutputStream. This explains why the User test with a large array is so much faster with the stream variant.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -1153,9 +1153,7 @@ public OutputStreamRandomAccessEncoder(RandomAccessOutputStream out) {
       void writeUInt32Field(int fieldNumber, int value) throws IOException {
          int pos = out.getPosition();
          int tag = WireType.makeTag(fieldNumber, WireType.WIRETYPE_VARINT);
-         int tagLen = varIntBytes(tag);
-         int valueLen = varIntBytes(value);
-         out.ensureCapacity(pos + tagLen + valueLen);
+         out.ensureCapacity(pos + 5 + 5);
```
ryanemerson, line 1157: See other comment. I don't think we can assume 5 bytes for the tag as it can be negative, I think it should be fine for the value though.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -1166,9 +1164,7 @@ void writeUInt32Field(int fieldNumber, int value) throws IOException {
       void writeUInt64Field(int fieldNumber, long value) throws IOException {
          int pos = out.getPosition();
          int tag = WireType.makeTag(fieldNumber, WireType.WIRETYPE_VARINT);
-         int tagLen = varIntBytes(tag);
-         int valueLen = varIntBytes(value);
-         out.ensureCapacity(pos + tagLen + valueLen);
+         out.ensureCapacity(pos + 5 + 5);
```
ryanemerson, line 1168: I don't think we can assume 5 bytes for the  value here as `Long.MAX_VALUE` requires 9 bytes.

ryanemerson (changes requested):

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -1399,47 +1397,53 @@ private int writeVarInt64Direct(int i, long value) throws IOException {
...
-               }
-            };
-         }
-         // This just asserts every childEncoder is properly closed
-         assert childEncoder.fieldNumber == -1;
-         childEncoder.fieldNumber = number;
-         return childEncoder;
+      Encoder subEncoder(int number) throws IOException {
+         // We assume worse case for tag and 1 byte for length
+         out.ensureCapacity(6);
```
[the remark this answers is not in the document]
wburns: Yes, it was a mistake on my part that `varIntBytes` is actually `varLongBytes`. Guessing that is just how protobuf named them, but I will update the affected places.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -429,6 +571,66 @@ void writeFixed32(int value) {
...
+         int count = s.length();
+         for (i = 0; i < count; i++) {
+            c = s.charAt(i);
+            if (c > 127) {
+               // TODO: do this without allocating the byte[]
+               count = s.getBytes(StandardCharsets.UTF_8).length;
+               break;
+            }
+         }
+         writeVarint32(utf8String(count));
```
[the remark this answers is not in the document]
wburns: Yeah, I had other changes I reverted but missed this method. It was working fine though :)

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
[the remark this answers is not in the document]
wburns: Actually this case it is just a varInt, up to 5 bytes as the tag is only an int and not a long.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
wburns: Same here the tag is only an int so its max size is 5.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
wburns: Agreed, but it can be up to 10 bytes for a "variable long" afaik.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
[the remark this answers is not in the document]
ryanemerson: The field `number` is always a positive int, which must be between 1 and 536870911, but I think the computed tag value can be negative:

```
WireType.makeTag(536870911, WireType.WIRETYPE_LENGTH_DELIMITED) = -6
WireType.makeTag(536870911, WireType.WIRETYPE_VARINT) = -8
```

core/src/main/java/org/infinispan/protostream/annotations/impl/GeneratedMarshallerBase.java
```diff
@@ -46,30 +44,12 @@ protected final <T> void writeNestedMessage(BaseMarshallerDelegate<T> marshaller
...
       }
-
-      public TagWriterImpl getWriter() {
-         return writer;
-      }
-
-      public void close() throws IOException {
-         ctx.getWriter().writeBytes(fieldNumber, baos.getByteBuffer());
+      try (TagWriter nestedWriter = ((TagWriter) ctx).subWriter(fieldNumber, true)) {
+         marshallerDelegate.marshall((ProtobufTagMarshaller.WriteContext) nestedWriter, null, message);
```
pruivo, line 52: `ctx`does not need to implement `TagWriter`. You can do
```java
      try (TagWriter nestedWriter = ctx.getWriter().subWriter(fieldNumber, true)) {
         marshallerDelegate.marshall((ProtobufTagMarshaller.WriteContext) nestedWriter, null, message);
      }
```

processor/src/main/java/org/infinispan/protostream/annotations/impl/processor/MarshallerSourceCodeGenerator.java
```diff
@@ -176,6 +176,7 @@ private void generateMessageMarshaller(ProtoMessageTypeMetadata pmtm) throws IOE
          iw.printf("import %s;\n", toImportName != null ? toImportName : toImport.getName());
          iw.println();
       }
+      iw.print("import org.infinispan.protostream.*;");
```
pruivo, line 179: the import is not required if using the `var` keyword.

core/src/main/java/org/infinispan/protostream/annotations/impl/AbstractMarshallerCodeGenerator.java
```diff
@@ -778,14 +778,12 @@ protected void generateWriteMethodBody(IndentWriter iw, ProtoMessageTypeMetadata
                         mapFieldMetadata.getValue().getJavaTypeName(),
                         v
                   );
-                  iw.printf("try (NestedWriter $n = new NestedWriter($1, %d)) {\n", fieldMetadata.getNumber());
+                  iw.printf("try (TagWriter $n = $out.subWriter(%d, true)) {\n", fieldMetadata.getNumber());
```
pruivo, line 781:
```suggestion
                  iw.printf("try (var $n = $out.subWriter(%d, true)) {\n", fieldMetadata.getNumber());
```

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -362,15 +391,128 @@ void writeUTF8Field(int fieldNumber, String value) throws IOException {
...
+            return new WrappedEncoder(new OutputStreamRandomAccessEncoder(raos)) {
+               @Override
+               void close() throws IOException {
+                  Encoder.this.writeBytes(number, raos.getByteBuffer());
+                  super.close();
+               }
+            };
+         }
+      }
+   }
```
pruivo, line 406: I'm confused about how this is supposed to work. When the `subEncoder` returns, the `close` method will be invoked in `RandomAccessOutputStream`. Should `raos.close` be invoked inside the `WrappedEncoder.close` method? 
Or just remove the try-with-resources.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -1191,13 +1382,82 @@ private int writeVarInt64Direct(int i, long value) throws IOException {
...
+      Encoder subEncoder(int number) throws IOException {
+         out.ensureCapacity(6);
+         int pos = out.getPosition();
+         pos = writeVarInt32Direct(pos, WireType.makeTag(number, WireType.WIRETYPE_LENGTH_DELIMITED));
+         out.setPosition(pos + 1);
+         if (positions == null) {
+            positions = new int[10];
+         } else if (head == positions.length) {
+            int newSize = head + 10;
+            if (newSize > 50) {
```
pruivo, line 1436: The max nested is configurable in `org.infinispan.protostream.config.Configuration#maxNestedMessageDepth()`
This check can be removed IMO.

pruivo (changes requested):

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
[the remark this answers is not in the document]
wburns: Yeah, which is fine, but tag is a Java `int` which fits in 5 bytes.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -960,9 +1161,7 @@ void writeUInt32Field(int fieldNumber, int value) throws IOException {
       void writeUInt64Field(int fieldNumber, long value) throws IOException {
          int pos = out.getPosition();
          int tag = WireType.makeTag(fieldNumber, WireType.WIRETYPE_VARINT);
-         int tagLen = varIntBytes(tag);
-         int valueLen = varIntBytes(value);
-         out.ensureCapacity(pos + tagLen + valueLen);
+         out.ensureCapacity(pos + 5 + 10);
```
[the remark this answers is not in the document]
wburns: I feel we need a MAX_VARLONG_SIZE and MAX_VARINT_SIZE to distinguish between values that are limited to an `int` (4 bytes) and `long` (8 bytes). The tag for example is only the former.

core/src/main/java/org/infinispan/protostream/annotations/impl/GeneratedMarshallerBase.java
wburns: Ah, I didn't notice that before 👍

processor/src/main/java/org/infinispan/protostream/annotations/impl/processor/MarshallerSourceCodeGenerator.java
wburns: Sounds good 👍

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
wburns: This is something I didn't look more closely into, but we don't have access to the config at this spot. The reason the check is needed is because this code uses the new `canReuseWriter` method and thus we only create a two TagWriterImpl instances per marshalling call (one for the outer and one for the child that handles all the nested messages). I had debated about passing down the max depth to the encoder, which I should probably do I guess.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
wburns: Honestly, this was an intermediate step and abuses the fact that all impls of RandomAccessOutputStream don't do anything on close. I can fix this commit. If you notice the last commit for this doesn't even invoke close anymore.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -241,6 +249,16 @@ public void writeRawBytes(byte[] value, int offset, int length) throws IOExcepti
       encoder.writeBytes(value, offset, length);
    }
 
+   @Override
+   public TagWriter subWriter(int number, boolean nested) throws IOException {
+      Encoder subEncoder = encoder.subEncoder(number);
```
pruivo, line 254: @wburns you can have the max nested depth here:
```suggestion
      Encoder subEncoder = encoder.subEncoder(number, serCtx.getConfiguration().maxNestedMessageDepth());
```

wburns: Actually, this is not where I wanted to add it, let me look into this a bit more again now that I have time :)

wburns: Ended up just implementing as an additional argument as I couldn't get the other ways to work the way I wanted.

wburns:
Updated, I will squash the commits later today if no more comments.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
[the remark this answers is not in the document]
ryanemerson: Argh I see now. When I changed the param type of `varIntBytes`  to int, I get 5 :+1:

ryanemerson:
> Updated, I will squash the commits later today if no more comments.

Looks good in general, but there are a lot of test failures according to the GH check.

wburns:
> > Updated, I will squash the commits later today if no more comments.
> 
> Looks good in general, but there are a lot of test failures according to the GH check.

Should be fixed now, when I added the maxDepth the arguments got switched in one case 🤦

core/src/main/java/org/infinispan/protostream/annotations/impl/GeneratedMarshallerBase.java
```diff
@@ -46,30 +44,12 @@ protected final <T> void writeNestedMessage(BaseMarshallerDelegate<T> marshaller
...
-         writeMessage(marshallerDelegate, nested.writer, message);
-      }
-   }
-
-   public static class NestedWriter implements AutoCloseable {
-      final RandomAccessOutputStream baos;
-      final TagWriterImpl writer;
-      private final ProtobufTagMarshaller.WriteContext ctx;
-      private final int fieldNumber;
+      if (!(ctx instanceof TagWriter)) {
```
pruivo, line 47: this check can be removed.

wburns:
Updated and addressed comments.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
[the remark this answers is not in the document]
pruivo: @wburns planning to replace all the code `5` with `MAX_INT_VARINT_SIZE` in this class?

pruivo (approved):
LGTM

wburns:
Yes, I meant to update them all. Guessing when I either missed or when I rebased I messed it up. I will fix them all and repush.

wburns:
Updated.

ryanemerson:
Thanks @wburns

PR #438, opened 2025-03-05
[#435] Always run source jar generation

### LINKED ISSUE #435: Source jars are missing
Since #423 , the generation of source jars is broken

tristantarrant:
Closes #435


ryanemerson:
Thanks @tristantarrant

PR #449, opened 2025-03-28
[#447] Utilise String coder and value field to optimise writing of ASCII strings

### LINKED ISSUE #447: Utilise String coder and value field to optimise writing of ASCII strings
It's possible to use a VarHandle to obtain the `coder` field on a String instance in order to determine whether the string uses only ASCII characters and if so, we can write the bytes directly from the String byte[]. Utilising a VarHandle for these fields requires that `--add-opens java.base/java.lang=ALL-UNNAMED` is provided as a JVM arg, therefore we should ensure that we can failback to the #413 optimisations when access to the coder is not possible.

ryanemerson:
Closes #447

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/impl/Log.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/Log.java
@@ -151,6 +151,10 @@ public interface Log extends BasicLogger {
    @Message(value = "Incompatible @ProtoTypeId in '%s', from '%s' to '%s'", id = 40)
    String incompatibleTypeIds(String name, Integer t1, Integer t2);
 
+   @LogMessage(level = WARN)
+   @Message(value = "Unable to utilize more performant reflection based String serialization. Please pass `--add-opens java.base/java.lang=ALL-UNNAMED` to the JVM to enable", id = 41)
+   void unableToRetrieveMethodHandles(@Cause Throwable t);
+
    class LogFactory {
       public static Log getLog(Class<?> clazz) {
          return Logger.getMessageLogger(Log.class, clazz.getName());
--- a/core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
@@ -1,0 +1,157 @@
+package org.infinispan.protostream.impl;
+
+import static java.lang.invoke.MethodType.methodType;
+
+import java.lang.invoke.CallSite;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import sun.misc.Unsafe;
+
+class StringUtil {
+
+   interface Utf8Helper {
+      boolean providesLatin1Bytes();
+      byte[] getBytes(String s);
+   }
+
+   private static final Log log = Log.LogFactory.getLog(StringUtil.class);
+
+   static final Utf8Helper UTF8_HELPER = getUtf8Helper();
+
+   private static Utf8Helper getUtf8Helper() {
+      // Attempt to use IMPL_LOOKUP so that we utilise a Lambda - Most efficient
+      try {
+         var helper = createLambdaHelper();
+         if (log.isDebugEnabled())
+            log.debug("Utilising Lambda based UNSAFE String optimisations");
+         return helper;
+      } catch (Throwable ignore) {
+      }
+
+      // Fallback to MethodHandler - Slightly less efficient and requires  --add-opens java.base/java.lang=ALL-UNNAMED
+      try {
+         return createFieldHelper();
+      } catch (Throwable t) {
+         log.unableToRetrieveMethodHandles(t);
+      }
+
+      // Coder optimisations not possible
+      return new Utf8Helper() {
+         @Override
+         public boolean providesLatin1Bytes() {
+            return false;
+         }
+
+         @Override
+         public byte[] getBytes(String s) {
+            return s.getBytes(StandardCharsets.UTF_8);
+         }
+      };
+   }
+
+   private static Utf8Helper createLambdaHelper() throws Throwable {
+      Class<MethodHandles.Lookup> lookupClass = MethodHandles.Lookup.class;
+      Field implLookup = lookupClass.getDeclaredField("IMPL_LOOKUP");
+
+      Field theUnsafeField = Unsafe.class.getDeclaredField("theUnsafe");
+      theUnsafeField.setAccessible(true);
+      Unsafe unsafe = (Unsafe) theUnsafeField.get(null);
+      long fieldOffset = unsafe.staticFieldOffset(implLookup);
+      MethodHandles.Lookup trustedLookup = (MethodHandles.Lookup) unsafe.getObject(lookupClass, fieldOffset);
+      MethodHandle constructor = trustedLookup.findConstructor(
+            MethodHandles.Lookup.class,
+            methodType(void.class, Class.class, Class.class, int.class)
+      );
+
+      MethodHandles.Lookup lookup = (MethodHandles.Lookup) constructor.invoke(String.class, null, -1);
+      MethodHandle isLatin1 = lookup.findSpecial(
+            String.class,
+            "isLatin1",
+            methodType(boolean.class),
+            String.class
+      );
+      CallSite predicate = LambdaMetafactory.metafactory(
+            lookup,
+            "test",
+            methodType(Predicate.class),
+            methodType(boolean.class, Object.class),
+            isLatin1,
+            MethodType.methodType(boolean.class, String.class)
+      );
+      Predicate<String> isLatin1Predicate = (Predicate<String>) predicate.getTarget().invokeExact();
+
+      MethodHandle value = lookup.findSpecial(
+            String.class,
+            "value",
+            methodType(byte[].class),
+            String.class
+      );
+      CallSite apply = LambdaMetafactory.metafactory(
+            lookup,
+            "apply",
+            methodType(Function.class),
+            methodType(Object.class, Object.class),
+            value,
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+            if (coderFn.applyAsInt(s) != LATIN1)
+               return s.getBytes(StandardCharsets.UTF_8);
+
+            return valueFn.apply(s);
+         }
+      };
+   }
+
+   private static Utf8Helper createFieldHelper() throws Throwable {
+      Method privateMethod = String.class.getDeclaredMethod("coder");
```
wburns, line 128: I suggest we use `isLatin1` instead. We aren't covering the case of when compact strings aren't used with this.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -309,6 +309,8 @@ public ProtoStreamWriterImpl getProtoStreamWriter() {
 
    //todo [anistor] need to provide a safety mechanism to limit message size in bytes and message nesting depth on write ops
    private abstract static class Encoder {
+
+      protected final StringUtil.Utf8Helper utf8Helper = StringUtil.getUtf8Helper();
```
wburns, line 313: We need to make this static, so the if block below can surely be optimized away as necessary.

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+         } catch (Throwable t) {
+            log.warnf(t, "Unable to retrieve MethodHandles required for Lambda based String optimisations");
+         }
+
+         // Fallback to MethodHandler - Slightly less efficient and requires  --add-opens java.base/java.lang=ALL-UNNAMED
+         try {
+            INSTANCE = createFieldHelper();
+            return INSTANCE;
+         } catch (Throwable t) {
+            log.warnf(t, "Unable to retrieve MethodHandles required for field based String optimisations");
```
wburns, line 45: Should we mention `--add-opens java.base/java.lang=ALL-UNNAMED` is required to enable the optimization?

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+   static volatile Utf8Helper INSTANCE;
+
+   static Utf8Helper getUtf8Helper() {
+      if (INSTANCE == null) {
+         // Attempt to use IMPL_LOOKUP so that we utilise a Lambda - Most efficient
+         try {
+            INSTANCE = createLambdaHelper();
+            return INSTANCE;
+         } catch (Throwable t) {
+            log.warnf(t, "Unable to retrieve MethodHandles required for Lambda based String optimisations");
```
wburns, line 37: Should these be WARN, is INFO not sufficient? Also I feel like we shouldn't log this if Unsafe isn't available as it will always be printed on startup and the user would have to downgrade to remove it.

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+   private static final Log log = Log.LogFactory.getLog(ProtoStreamReaderImpl.class);
+   private static final Byte LATIN1 = (byte) 0;
+
+   static volatile Utf8Helper INSTANCE;
+
+   static Utf8Helper getUtf8Helper() {
+      if (INSTANCE == null) {
+         // Attempt to use IMPL_LOOKUP so that we utilise a Lambda - Most efficient
+         try {
+            INSTANCE = createLambdaHelper();
```
wburns, line 34: I still question if this is valuable especially once Java 25 comes out, I am guessing many will _try_ to be on that version.

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
wburns: Also we need i18n for these messages.

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
wburns: Maybe we should add these to the commons all/jdk22 to better separate and not have to worry about checking for Unsafe? Then this code can also use `UnsafeHolder` to not have to repeat the code.

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+import sun.misc.Unsafe;
+
+class StringUtil {
+
+   interface Utf8Helper {
+      boolean providesLatin1Bytes();
+      byte[] getBytes(String s);
+   }
+
+   private static final Log log = Log.LogFactory.getLog(ProtoStreamReaderImpl.class);
```
wburns, line 25: Wrong class name.

wburns (changes requested):

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+   private static final Byte LATIN1 = (byte) 0;
+
+   static volatile Utf8Helper INSTANCE;
```
franz1981, line 28: I don't get why you cannot just have a `static final VarHandle CODER_FIELD` to use.
Since the JVM trust static final fields to treat them as constants, this is necessary to get the proper expected performance...

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
franz1981: static and final :pray:

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
```diff
@@ -0,0 +1,155 @@
...
+         public byte[] getBytes(String s) {
+            if (coderFn.applyAsInt(s) != LATIN1)
+               return s.getBytes(StandardCharsets.UTF_8);
+
+            return valueFn.apply(s);
+         }
+      };
+   }
+
+   private static Utf8Helper createFieldHelper() throws Throwable {
```
franz1981, line 127: Al the stuff here need to be static final or it will perform terribly - see https://shipilev.net/jvm/anatomy-quarks/17-trust-nonstatic-final-fields/

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
ryanemerson: Updated so the output is:

```java
Apr 08, 2025 4:36:26 PM org.infinispan.protostream.impl.StringUtil getUtf8Helper
WARN: IPROTO000041: Unable to retrieve MethodHandles required for field based String optimisations. `--add-opens java.base/java.lang=ALL-UNNAMED` must be passed to the JVM
java.lang.reflect.InaccessibleObjectException: Unable to make boolean java.lang.String.isLatin1() accessible: module java.base does not "opens java.lang" to unnamed module @21b8d17c
	at java.base/java.lang.reflect.AccessibleObject.throwInaccessibleObjectException(AccessibleObject.java:391)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:367)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:315)
	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:203)
	at java.base/java.lang.reflect.Method.setAccessible(Method.java:197)
	at org.infinispan.protostream.impl.StringUtil.createFieldHelper(StringUtil.java:127)
	at org.infinispan.protostream.impl.StringUtil.getUtf8Helper(StringUtil.java:43)
	at org.infinispan.protostream.impl.StringUtil.<clinit>(StringUtil.java:28)
```

core/src/main/java/org/infinispan/protostream/impl/StringUtil.java
ryanemerson: Updated so that we only output a WARN message if the `createFieldHelper` lookup fails.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
```diff
@@ -309,6 +309,8 @@ public ProtoStreamWriterImpl getProtoStreamWriter() {
 
    //todo [anistor] need to provide a safety mechanism to limit message size in bytes and message nesting depth on write ops
    private abstract static class Encoder {
+
+      protected final StringUtil.Utf8Helper utf8Helper = StringUtil.UTF8_HELPER;
```
wburns, line 313:
```suggestion
      protected static final StringUtil.Utf8Helper utf8Helper = StringUtil.UTF8_HELPER;
```

core/src/main/java/org/infinispan/protostream/impl/Log.java
```diff
@@ -151,6 +151,10 @@ default MalformedProtobufException messageTruncated() {
    @Message(value = "Incompatible @ProtoTypeId in '%s', from '%s' to '%s'", id = 40)
    String incompatibleTypeIds(String name, Integer t1, Integer t2);
 
+   @LogMessage(level = WARN)
+   @Message(value = "Unable to retrieve MethodHandles required for field based String optimisations. `--add-opens java.base/java.lang=ALL-UNNAMED` must be passed to the JVM", id = 41)
```
wburns, line 155: Wondering if we want to word this a bit differently, maybe something like?

```suggestion
   @Message(value = "Unable to utilize more performant reflection based String serialization. Please pass `--add-opens java.base/java.lang=ALL-UNNAMED` to the JVM to enable", id = 41)
```

wburns (commented):
LGTM, just some minor things.

core/src/main/java/org/infinispan/protostream/impl/TagWriterImpl.java
ryanemerson: Thanks, I thought I had done this! :see_no_evil:

wburns:
Integrated into main, thanks @ryanemerson !

PR #488, opened 2025-07-14
[#487] JSON serialization use only descriptor

### LINKED ISSUE #487: JSON serialization requires a marshaller
With the changes introduced by #405, the JSON serialization now requires the marshaller to be registered. We need to fix that so it can work as previously without this requirement. Having only the proto definition should be enough to transform a JSON into the byte representation.

jabolina:
* Do not require a marshaller to be registered.

Closes #487.


wburns:
Integrated into main, thanks @jabolina !

PR #524, opened 2025-08-27
[#522] WrappedMessage should avoid byte[] allocation for inner message

### LINKED ISSUE #522: WrappedMessage should avoid byte[] allocation for inner message
This byte[] allocation for a nested message https://github.com/infinispan/protostream/blob/30b95bf972a81d38b336589b4103b11945c46039/core/src/main/java/org/infinispan/protostream/WrappedMessage.java#L554 can use a ByteBuffer instead which should provide a constant allocation cost instead of a possibly very large byte[] if the nested message is bigger, which can very easily with any user key or value object references.

wburns:
Fixes #522

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/WrappedMessage.java
+++ b/core/src/main/java/org/infinispan/protostream/WrappedMessage.java
@@ -1,6 +1,7 @@
 package org.infinispan.protostream;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Date;
@@ -519,7 +520,7 @@ public final class WrappedMessage {
       String typeName = null;
       Integer typeId = null;
       int enumValue = -1;
-      byte[] messageBytes = null;
+      ByteBuffer messageBytes = null;
       Object value = null;
       int fieldCount = 0;
       int expectedFieldCount;
@@ -551,7 +552,7 @@ public final class WrappedMessage {
             }
             case WRAPPED_MESSAGE << WireType.TAG_TYPE_NUM_BITS | WireType.WIRETYPE_LENGTH_DELIMITED: {
                expectedFieldCount = 2;
-               messageBytes = in.readByteArray();
+               messageBytes = in.readByteBuffer();
                break;
             }
             case WRAPPED_INSTANT_SECONDS << WireType.TAG_TYPE_NUM_BITS | WireType.WIRETYPE_VARINT: {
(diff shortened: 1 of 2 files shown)


wburns:
Performance results from running https://github.com/infinispan/infinispan-benchmarks/pull/28

MAIN:

Benchmark | (useUserObjectValue) | (userByteArraySize) | Mode | Cnt | Score |   | Error | Units
-- | -- | -- | -- | -- | -- | -- | -- | --
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | FALSE | 10 | avgt | 10 | 200.259 | ± | 0.507 | ns/op
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | FALSE | 8096 | avgt | 10 | 1048.549 | ± | 3.603 | ns/op
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | TRUE | 10 | avgt | 10 | 626.33 | ± | 17.285 | ns/op
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | TRUE | 8096 | avgt | 10 | 1929.314 | ± | 4.102 | ns/op

CHANGES:


Benchmark | (useUserObjectValue) | (userByteArraySize) | Mode | Cnt | Score |   | Error | Units
-- | -- | -- | -- | -- | -- | -- | -- | --
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | FALSE | 10 | avgt | 10 | 206.716 | ± | 0.891 | ns/op
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | FALSE | 8096 | avgt | 10 | 539.997 | ± | 3.213 | ns/op
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | TRUE | 10 | avgt | 10 | 594.386 | ± | 1.397 | ns/op
MarshallerBenchmark.testPutKeyValueCommandFromByteBuffer | TRUE | 8096 | avgt | 10 | 953.237 | ± | 4.691 | ns/op

So everything but the put key value command with a small array is faster. This is due to the overhead of a byte[] with just 10 bytes vs a HeapByteBuffer. The loss in the small byte[] case is worth it for larger array sizes though.

pruivo:
Can't the following place be improved as well?

https://github.com/infinispan/protostream/blob/30b95bf972a81d38b336589b4103b11945c46039/core/src/main/java/org/infinispan/protostream/WrappedMessage.java#L631

https://github.com/infinispan/protostream/blob/30b95bf972a81d38b336589b4103b11945c46039/core/src/main/java/org/infinispan/protostream/WrappedMessage.java#L719

wburns:
It looks like ProtobufParser#parseMessage could also be refactored, but it has a few layers of abstractions around the method calls, so I don't think it is worth it.

pruivo (approved):

ryanemerson (approved):

PR #525, opened 2025-08-27
[#523] We should be able to optimize message id lookup

### LINKED ISSUE #523: We should be able to optimize message id lookup
Currently when looking up a nested message marshaller by id we first lookup the type name and then use the type name to lookup the marshaller. Both of these are HashMap lookup calls. https://github.com/infinispan/protostream/blob/30b95bf972a81d38b336589b4103b11945c46039/core/src/main/java/org/infinispan/protostream/WrappedMessage.java#L590-L593

There are two things here that should help:

1. Condense the id -> String -> marshaller lookup to be just int -> marshaller to remove a HashMap lookup
2. We can also add a custom Map implementation that stores ids with values less than say 4096 in an array along with the HashMap so that very frequent ids are just array lookups instead of a HashMap lookup.

With both of these changes we would change two HashMap lookups into a single array access which is considerably faster.

wburns:
* Condense id -> typeName -> marshaller to id -> marshaller lookup
* Add array based id optimization for small ids

Fixes #523

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/descriptors/ResolutionContext.java
+++ b/core/src/main/java/org/infinispan/protostream/descriptors/ResolutionContext.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import org.infinispan.protostream.DescriptorParserException;
 import org.infinispan.protostream.FileDescriptorSource;
 import org.infinispan.protostream.impl.Log;
+import org.infinispan.protostream.impl.SmallIntMap;
 
 /**
  * @author anistor@redhat.com
@@ -22,7 +23,7 @@ public class ResolutionContext {
 
    private final Map<String, FileDescriptor> fileDescriptorMap;
 
-   private final Map<Integer, GenericDescriptor> allTypeIds;
+   private final SmallIntMap<GenericDescriptor> allTypeIds;
 
    private final Map<Integer, GenericDescriptor> typeIds = new HashMap<>();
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/infinispan/protostream/impl/SerializationContextImpl.java
```diff
@@ -492,6 +521,32 @@ public <T> BaseMarshallerDelegate<T> getMarshallerDelegate(String typeName) {
...
+      // Try first without locking as this class is thread safe
+      Registration registration = marshallersByTypeId.get(typeId);
+      if (registration != null) {
+         return (BaseMarshallerDelegate<T>) registration.marshallerDelegate;
+      }
+
+      // If the id wasn't found we have to acquire the read lock to ensure no concurrent updates
+      long stamp = manifestLock.readLock();
+      try {
+         registration = marshallersByTypeId.get(typeId);
```
pruivo, line 535: Why do you need to do the lookup in `marshallersByTypeId` again?

wburns: It was for concurrent updates, but now that I made the SmallIntMap fully thread safe (to avoid the lock completely), it _probably_ isn't needed. I will mull over it a bit and update.

core/src/main/java/org/infinispan/protostream/descriptors/ResolutionContext.java
```diff
@@ -155,9 +156,14 @@ private <K, V> V lookup(Map<K, V> first, Map<K, V> second, K k) {
...
 
+   private <V> V lookup(Map<Integer, V> first, SmallIntMap<V> second, int k) {
+      V v = first.get(k);
+      return v != null ? v : second.get(k);
+   }
+
    void flush() {
       allGlobalTypes.putAll(globalTypes);
-      allTypeIds.putAll(typeIds);
+      typeIds.forEach(allTypeIds::put);
```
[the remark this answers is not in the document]
wburns: Correct, afaik, it is only done at startup. I can add the method though

pruivo (approved):

pruivo:
merged! thanks @wburns !

PR #539, opened 2025-09-09
[#538] InputStream based unmarshalling via ProtobufUtil is quite slow

### LINKED ISSUE #538: InputStream based unmarshalling via ProtobufUtil is quite slow
When comparing a byte backed and InputStream unsmarshalling it is found that the latter can be three times slower when using a small message payload. A larger message is also slower but not by such a degree.

We need a more optimized version. Most likely we can add a buffered decoder variant that reads up to say 128 bytes at a time to avoid large allocations and smaller entries should be similar performance to the byte backed, just a tad slower as it has to allocate the byte[]. Although maybe these can be removed.

wburns:
Fixes #538


core/src/main/java/org/infinispan/protostream/impl/RandomAccessOutputStreamImpl.java
```diff
@@ -17,8 +17,8 @@ public class RandomAccessOutputStreamImpl extends OutputStream implements Random
 
    static final int DEFAULT_DOUBLING_SIZE = 4 * 1024 * 1024; // 4MB
 
-   byte[] buf;
-   int pos = 0;
+   protected byte[] buf;
+   protected int pos = 0;
```
ryanemerson, line 21: Why the change, AFAICT these variables are only used in this class.

PR #541, opened 2025-09-09
[#540] Add a no op encoder that over estimates size

### LINKED ISSUE #540: Add a no op encoder that over estimates size
We have a NoOpEncoder today that gives an exact size of a marshalled object. It might be useful to have a faster implementation that does a size estimation that will never be less than the actual size. For example JGroups needs this for its size operations. This allows our calls to have no looping and be constant time for the actual "write" portions.

wburns:
Fixes #540

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
+++ b/core/src/main/java/org/infinispan/protostream/ProtobufUtil.java
@@ -55,8 +55,20 @@ public final class ProtobufUtil {
       return out.getWrittenBytes();
    }
 
+   public static <A> int estimateSize(ImmutableSerializationContext ctx, A t) throws IOException {
+      TagWriterImpl out = TagWriterImpl.newInstanceSizeEstimate(ctx);
+      write(ctx, out, t);
+      return out.getWrittenBytes();
+   }
+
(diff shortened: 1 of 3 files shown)


wburns:
Benchmark from https://github.com/infinispan/infinispan-benchmarks/pull/30 as you can see it is faster in every case expected. The ones with nested messages were especially faster.


| Benchmark | byteArrayOrStream | userByteArraySize | Score | Units |
| :--- | :--- | :--- | :--- | :--- |
| `testSizeEstimateAddress` | FALSE | N/A | 25.067 | ns/op |
| `testSizedAddress` | FALSE | N/A | 28.041 | ns/op |
| `testSizeEstimateIracMetadata` | FALSE | N/A | 35.034 | ns/op |
| `testSizedIracMetadata` | FALSE | N/A | 53.697 | ns/op |
| `testSizeEstimateUUID` | FALSE | N/A | 19.458 | ns/op |
| `testSizedUUID` | FALSE | N/A | 21.300 | ns/op |
| `testSizeEstimateUser` | FALSE | 10 | 53.630 | ns/op |
| `testSizedUser` | FALSE | 10 | 118.592 | ns/op |
| `testSizeEstimateUser` | FALSE | 8096 | 53.898 | ns/op |
| `testSizedUser` | FALSE | 8096 | 120.505 | ns/op |

PR #582, opened 2025-11-11
[#581] Fix records as valid outer elements

### LINKED ISSUE #581: Inner elements of records are orphans
When calculating a type's outer element, records are ignored, causing inner types to be orphans. This causes issues with calculating a type's name, especially troublesome when migrating classes to records.

For example:

```java
class Outer {
   public enum Inner {}
}
```
qualifies the inner type as `Outer.Inner`, whereas

```java
record Outer() {
   public enum Inner {}
}
```
qualifies the inner type as `Inner`.

tristantarrant:
Closes #581


processor-tests/src/test/java/org/infinispan/protostream/processor/tests/ProtoSchemaTest.java
```diff
@@ -169,6 +171,8 @@ public void testGeneratedInitializer() throws Exception {
       ProtobufUtil.toWrappedByteArray(ctx, new Note());
 
       assertTrue(ctx.canMarshall(SimpleClass.class));
+      assertTrue(ctx.canMarshall(OuterRecord.class));
+      assertTrue(ctx.canMarshall("firstTestPackage.OuterRecord.InnerEnum"));
```
wburns: Just for completeness can we add `assertFalse(ctx.canMarshall("firstTestPackage.InnerEnum"));` just as well? This should then fail without the change right?

PR #670, opened 2026-05-01
[#669] Copy-on-write serialization context state

### LINKED ISSUE #669: copy-on-write for serialization context state
In the serialization context, for thread-safety, we utilize read-write locks to guard updates and reads of the marshallers and descriptors registered. Although it is a read-write lock, the implementation still needs to perform CAS operations to acquire and release the read lock, and this is in the hot path for serialization.

Since the registration/removal/updates of the context are infrequent, we could utilize a COW approach and remove the lock contention of the read path and replace it with a volatile read.

jabolina:
* Replace the rw-locks in the serialization context with a COW structure.
* A lock is held for atomic and consistent writes. Readers only perform a volatile read.

I also updated the `SmallIntMap` to use the same approach. This assumes that reads are more frequent than writes, too.

Here is some data. I run the tests once with `-t 1`, so no concurrent contention, and again with `-t 16` to have concurrent access:

| Benchmark | Payload | Baseline (t=1) | COW (t=1) | Delta | Baseline (t=16) | COW (t=16) | Delta |                                                                                                                                         
|---|---|---|---|---|---|---|---|
| serialize | 10B | 231.6 | 208.3 | -10.1% | 5,761 | 600 | **-89.6%** |                                                                                                                                                                     
| serialize | 8KB | 1,279 | 1,233 | -3.6% | 6,598 | 4,099 | **-37.9%** |                                                                                                                                                                    
| deserialize | 10B | 325.9 | 309.3 | -5.1% | 2,575 | 800 | **-68.9%** |                                                                                                                                                                    
| deserialize | 8KB | 892.6 | 869.3 | -2.6% | 4,182 | 3,774 | **-9.8%** | 

Running with perf, we can see the improvements for instruction/cycle under concurrent access (`-t 16`):

| Benchmark | Payload | Metric | Baseline | COW |                                                                                                                                                                                           
|---|---|---|---|---|
| serialize | 10B | cycles (↓) | 18,291 | 1,283 |                                                                                                                                                                                           
| serialize | 10B | instructions (↓) | 7,742 | 5,848 |                                                                                                                                                                                      
| serialize | 10B | **IPC (↑)** | **0.42** | **4.56** |
| serialize | 8KB | cycles (↓) | 19,277 | 11,702 |                                                                                                                                                                                          
| serialize | 8KB | **IPC (↑)** | **0.56** | **0.84** |
| deserialize | 10B | cycles (↓) | 9,718 | 2,014 |                                                                                                                                                                                          
| deserialize | 10B | **IPC (↑)** | **1.26** | **3.76** |
| deserialize | 8KB | cycles (↓) | 11,019 | 10,709 |                                                                                                                                                                                        
| deserialize | 8KB | **IPC (↑)** | **0.95** | **0.89** | 

↓: lower is better
↑: higher is better

This was running with the `marshallher-benchmark` from `infinispan-benchmarks`:

```java
    @Benchmark
    public ByteBuffer testUserToByteBuffer(InfinispanHolder holder, UserState userState) throws Exception {
        return holder.getMarshaller().objectToBuffer(userState.getUser());
    }

    @Benchmark
    public Object testUserFromByteBuffer(InfinispanHolder holder, UserState userState) throws Exception {
       ByteBuffer bb = userState.getUserBytes();
        return holder.getMarshaller().objectFromByteBuffer(bb.getBuf(), bb.getOffset(), bb.getLength());
    }
```

Closes #669

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/impl/SmallIntMap.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/SmallIntMap.java
@@ -3,7 +3,7 @@ package org.infinispan.protostream.impl;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.locks.StampedLock;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * A map-like data structure that uses a copy-on-write array for keys up to a certain threshold 
@@ -16,17 +16,22 @@ public class SmallIntMap<V> {
 
    private static final int DEFAULT_ARRAY_THRESHOLD = 4096;
 
-   private volatile V[] array;
-   private Map<Integer, V> map;
+   private volatile State<V> state;
 
-   private final StampedLock lock = new StampedLock();
+   private final ReentrantLock lock = new ReentrantLock();
 
    public SmallIntMap() {
       this(DEFAULT_ARRAY_THRESHOLD);
    }
 
    public SmallIntMap(int arrayThreshold) {
-      this.array = (V[]) new Object[arrayThreshold];
+      this.state = new State<>(allocate(arrayThreshold), null);
+   }
+
+   public SmallIntMap(SmallIntMap<V> other) {
+      V[] arr = Arrays.copyOf(other.state.array, other.state.array.length);
+      Map<Integer, V> map = other.state.map != null ? new HashMap<>(other.state.map) : null;
+      this.state = new State<>(arr, map);
    }
 
    /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/infinispan/protostream/impl/SmallIntMap.java
```diff
@@ -16,17 +16,21 @@ public class SmallIntMap<V> {
 
    private static final int DEFAULT_ARRAY_THRESHOLD = 4096;
 
-   private volatile V[] array;
-   private Map<Integer, V> map;
-
-   private final StampedLock lock = new StampedLock();
+   private final ReentrantLock writeLock = new ReentrantLock();
```
wburns: Any reason we don't just use the StampedLock still? It can do ReentrantLock operations and performance before was tested to be better, even when used for write/read lock operations.

jabolina: Sounds good, I can update.

jabolina: Updated.

wburns: Sorry, I misread the changes ReentrantLock is the preferred choice here.

core/src/main/java/org/infinispan/protostream/impl/SmallIntMap.java
```diff
@@ -122,39 +117,56 @@ public V get(int key) {
...
             newArray[key] = null;
-            this.array = newArray;
+            state = new State<>(newArray, s.map);
             return oldValue;
          }
-         if (map == null) {
-            return null;
-         }
-         return map.remove(key);
+         Map<Integer, V> newMap = s.map != null ? new HashMap<>(s.map) : new HashMap<>();
```
wburns, line 133: No reason to do `new HashMap` if map is null, we could just do `if (s.map == null) { return null; }` like the code was before. But no reason to worry as it shouldn't really affect our usage.

wburns (approved):

wburns:
Integrated into main, thanks @jabolina !

PR #680, opened 2026-05-27
Fix copy-on-write resolution losing types when other files fail to resolve

### LINKED ISSUE #679: Fix copy-on-write resolution losing types when other files fail to resolve
## Summary

- Re-add per-file `flush()` in `FileDescriptor.resolveDependencies()` so successfully resolved types are committed to the global maps immediately, preventing a subsequent file's failure from wiping them via `clear()`
- Remove `Collections.unmodifiableMap()` wrappers in `SerializationContextImpl.registerProtoFiles()` so `flush()` can write directly to the local copy maps
- Remove redundant `putAll(getResolved*())` calls since `flush()` handles this
- Add reproducer test `testResolvedTypesPreservedWhenOtherFileFails`

## Context

PR #670 introduced copy-on-write snapshots for `SerializationContext` state but removed the per-file `resolutionContext.flush()` call from `FileDescriptor.resolveDependencies()`. Without it, types from all resolved files accumulate in `ResolutionContext`'s local maps. When a later file fails to resolve, the `finally` block calls `clear()`, wiping all accumulated types — including those from previously successful files.

This surfaces in Infinispan's `ProtobufResourceTest` where schemas are removed one-by-one (leaving dependants unresolved), then new schemas are registered. The unresolved dependants fail to re-resolve during `resolve()`, and `clear()` destroys the newly registered schema's types.

## Test plan

- [ ] New test `testResolvedTypesPreservedWhenOtherFileFails` registers a valid and invalid file together, asserts the valid file's types survive
- [ ] All existing `SerializationContextImplTest` tests pass

wburns:
Fixes #679

## Summary

- Re-add per-file `flush()` in `FileDescriptor.resolveDependencies()` so successfully resolved types are committed to the global maps immediately, preventing a subsequent file's failure from wiping them via `clear()`
- Remove `Collections.unmodifiableMap()` wrappers in `SerializationContextImpl.registerProtoFiles()` so `flush()` can write directly to the local copy maps
- Remove redundant `putAll(getResolved*())` calls since `flush()` handles this
- Add reproducer test `testResolvedTypesPreservedWhenOtherFileFails`

## Context

PR #670 introduced copy-on-write snapshots for `SerializationContext` state but removed the per-file `resolutionContext.flush()` call from `FileDescriptor.resolveDependencies()`. Without it, types from all resolved files accumulate in `ResolutionContext`'s local maps. When a later file fails to resolve, the `finally` block calls `clear()`, wiping all accumulated types — including those from previously successful files.

This surfaces in Infinispan's `ProtobufResourceTest` where schemas are removed one-by-one (leaving dependants unresolved), then new schemas are registered. The unresolved dependants fail to re-resolve during `resolve()`, and `clear()` destroys the newly registered schema's types.

## Test plan

- [x] New test `testResolvedTypesPreservedWhenOtherFileFails` registers a valid and invalid file together, asserts the valid file's types survive
- [x] All existing `SerializationContextImplTest` tests pass

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/descriptors/FileDescriptor.java
+++ b/core/src/main/java/org/infinispan/protostream/descriptors/FileDescriptor.java
@@ -239,6 +239,7 @@ public final class FileDescriptor {
          }
 
          status = Status.RESOLVED;
+         resolutionContext.flush();
          resolutionContext.handleSuccess(this);
       } catch (DescriptorParserException dpe) {
          resolutionContext.handleError(this, dpe);
--- a/core/src/main/java/org/infinispan/protostream/descriptors/ResolutionContext.java
+++ b/core/src/main/java/org/infinispan/protostream/descriptors/ResolutionContext.java
@@ -161,16 +161,11 @@ public class ResolutionContext {
       return v != null ? v : second.get(k);
    }
 
-   public Map<String, GenericDescriptor> getResolvedTypes() {
-      return globalTypes;
-   }
-
-   public Map<String, EnumValueDescriptor> getResolvedEnumValues() {
-      return enumValueDescriptors;
-   }
-
-   public Map<Integer, GenericDescriptor> getResolvedTypeIds() {
-      return typeIds;
+   void flush() {
+      allGlobalTypes.putAll(globalTypes);
+      allTypeIds.putAll(typeIds);
+      allEnumValueDescriptors.putAll(enumValueDescriptors);
+      clear();
    }
 
    void clear() {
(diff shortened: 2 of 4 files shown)


wburns:
Talking with @jabolina it would be better to change this to have a context that is immutable and then only provider builders is some capacity instead of what we are currently doing, but we need to ensure public API compatibility which this PR fixes for the current iteration.

tristantarrant (approved):
Compiled, tested and installed locally and ran the Infinispan testsuite with it and it passes.

PR #688, opened 2026-06-05
Update entity first steps

karesti:
Start https://github.com/infinispan/protostream/issues/712

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/ProtobufFieldUpdater.java
+++ b/core/src/main/java/org/infinispan/protostream/ProtobufFieldUpdater.java
@@ -1,0 +1,274 @@
+package org.infinispan.protostream;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.infinispan.protostream.descriptors.Descriptor;
+import org.infinispan.protostream.descriptors.FieldDescriptor;
+import org.infinispan.protostream.descriptors.Type;
+import org.infinispan.protostream.impl.Log;
+import org.infinispan.protostream.impl.RandomAccessOutputStreamImpl;
+import org.infinispan.protostream.impl.TagWriterImpl;
+
+/**
+ * Partially updates protobuf-serialized message bytes at the field level, without requiring Java POJOs.
+ * Supports scalar field SET (including nested dotted paths), and collection ADD/REMOVE on repeated fields.
+ *
+ * @since 6.0
+ */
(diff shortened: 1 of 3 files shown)


tristantarrant:
This needs an issue, and possibly a link to the overarching epic

core/src/main/java/org/infinispan/protostream/ProtobufFieldUpdater.java
```diff
@@ -0,0 +1,331 @@
...
+         applyOperation(descriptor, fieldMap, op.type(), op.propertyPath(), 0, op.values());
+      }
+
+      return serializeFields(descriptor, fieldMap);
+   }
+
+   private static Map<Integer, List<Object>> collectFields(Descriptor descriptor, byte[] bytes) throws IOException {
+      Map<Integer, List<Object>> fieldMap = new LinkedHashMap<>();
+
+      ProtobufParser.INSTANCE.parse(new TagHandler() {
```
tristantarrant: The anonymous TagHandler should be replaced by a `FieldCollector` instance passed to `ProtobufParser.INSTANCE.parse()`, since `FieldCollector`.
This would simplify `collectFields` to 
```java
      FieldCollector collector = new FieldCollector();
      ProtobufParser.INSTANCE.parse(collector, descriptor, bytes);
      return collector.fields;
```

core/src/main/java/org/infinispan/protostream/ProtobufFieldUpdater.java
```diff
@@ -0,0 +1,331 @@
...
+            }
+         }
+      }
+
+      writer.flush();
+      return baos.toByteArray();
+   }
+
+   private static void writeField(TagWriter writer, int fieldNumber, FieldDescriptor fd, Object value) throws IOException {
+      if (fd == null) {
```
tristantarrant, line 198: This is tricky: if we don't have the field descriptor it means we are dealing with an older schema, and this code is making assumptions about the actual protobuf field type based simply on the java type of the value.
I think this scenario might only happen if the schema is out-of-date with what the client is sending. I've created https://github.com/infinispan/infinispan/issues/17681 to address that situation (the server should never store data incompatible with the schema for that cache).
In the meantime this should be a `throw Log.LOG.unknownField()` as an `IllegalStateException`

tristantarrant (changes requested):
Also, please remove the AI-CODE.md file (that should be handled in a dedicated PR),
create an issue and link it to the overarching Epic for this feature

karesti:
@tristantarrant I updated

tristantarrant (approved):

tristantarrant:
Merged, thanks @karesti

PR #707, opened 2026-06-22
[#706] Tighten GitHub Actions workflow permissions to least privilege

### LINKED ISSUE #706: Tighten GitHub Actions workflow permissions to least privilege
Several workflows grant broader GITHUB_TOKEN permissions than necessary:

- `ai_ide_config_notice.yml` has no top-level `permissions` block, so new jobs would inherit the default (read-all) instead of nothing
- `release.yml` `published` job inherits `contents: read` but only polls a Maven Central URL — needs no GitHub API access
- `test_report.yaml` `report-summary` job has `statuses: write` but only creates check runs (needs `checks: write`, not `statuses: write`)
- `add_issues_to_major_project.yaml` has top-level `contents: read` but only uses a PAT, never GITHUB_TOKEN

All action versions are already pinned to commit SHAs.

tristantarrant:
## Summary
- Add top-level `permissions: {}` to `ai_ide_config_notice.yml` so new jobs default to no permissions
- Add `permissions: {}` to `release.yml` `published` job (only polls Maven Central, no GitHub API)
- Remove unused `statuses: write` from `test_report.yaml` `report-summary` job (only uses `checks: write`)
- Change `add_issues_to_major_project.yaml` top-level permissions to `{}` (uses PAT, not GITHUB_TOKEN)

Closes #706

Created with the assistance of an AI tool

### DIFF
--- a/.github/workflows/add_issues_to_major_project.yaml
+++ b/.github/workflows/add_issues_to_major_project.yaml
@@ -8,8 +8,7 @@ on:
       - opened
       # Triggers the workflow when a new issue is opened in the repository.
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   add-to-project:
--- a/.github/workflows/ai_ide_config_notice.yml
+++ b/.github/workflows/ai_ide_config_notice.yml
@@ -28,6 +28,8 @@ on:
       - '**/AI.md'
       - '**/AI-*.md'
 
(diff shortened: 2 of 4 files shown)


jabolina (approved):

jabolina:
Thanks, @tristantarrant!

PR #739, opened 2026-09-08
[#738] Fix JSON serialization for nested maps and repeated fields

### LINKED ISSUE #738: JSON conversion fails for maps with nested object values and repeated fields as first field
## Problem

`ProtobufUtil.toCanonicalJSON` throws `IllegalStateException: Maps only have nested objects for values` when a map's value is a message that contains nested structures (sub-maps, repeated fields). This was introduced in PR #411 which rewrote the JSON writer hierarchy.

Additionally, `ArrayJsonWriter` drops the property name when a repeated field is the first content in a non-complex object (no `_type` written), producing malformed JSON.

## Root Cause

**MapJsonWriter:**
- `onTag` treated field number 1 as a map key even when the delegate (value message) was active
- `onStartNested` threw instead of forwarding to the delegate when the value message was open
- `onEndNested` cleared the delegate too early

**ArrayJsonWriter:**
- `onStartNested` only handled the `followedByComma(lastToken())` case but not `isOpen(lastToken())`, so the property name was missing when a repeated field is the first content in an object

## Fix

- `MapJsonWriter`: added `valueField` tracking so nested structures inside a map's message value are correctly delegated
- `ArrayJsonWriter`: added the `isOpen(lastToken())` branch to write the property name when the repeated field is first in an object

## Test plan

- `testMapWithNestedObjectValues` — mirrors the Keycloak `authenticationSessions` scenario
- `testMapWithMultipleNestedObjectValues` — multiple map entries with nested message values
- `testRepeatedFieldFirstInObject` — repeated field as first content in a non-complex nested object

Created with the assistance of an AI tool

tristantarrant:
Fixes #738

- MapJsonWriter: delegate nested structures inside map value messages instead of throwing IllegalStateException
- ArrayJsonWriter: write property name when repeated field is first in an object (lastToken is open brace)

Created with the assistance of an AI tool

### DIFF
--- a/core/src/main/java/org/infinispan/protostream/impl/json/ArrayJsonWriter.java
+++ b/core/src/main/java/org/infinispan/protostream/impl/json/ArrayJsonWriter.java
@@ -68,12 +68,15 @@ final class ArrayJsonWriter extends BaseJsonWriter {
          done = false;
       } else {
          // Otherwise, we are starting a fresh list.
-         // We populate the property name if we're starting a list where it is still missing.
-         if (JsonToken.followedByComma(lastToken())) {
-            pushToken(JsonToken.COMMA);
-            pushToken(JsonTokenWriter.string(fieldDescriptor.getName()));
-            pushToken(JsonToken.COLON);
-         }
+          // We populate the property name if we're starting a list where it is still missing.
+          if (JsonToken.followedByComma(lastToken())) {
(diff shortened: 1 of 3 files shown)


jabolina (approved):

jabolina:
Thanks, @tristantarrant!

